PR #1, opened 2020-07-16
added sap_system field [RHCLOUD-7696]

ryandillinfelton:
reflection of change to system profile in inventory here: https://github.com/RedHatInsights/insights-host-inventory/pull/679


schemas/system_profile/v1.yaml
```diff
@@ -198,3 +198,5 @@ $defs:
         type: array
         items:
           type: string
+      sap_system:
```
jharting, line 201: @ryandillinfelton can you add a bit of verbiage into the description field? Something along the lines of "Indicates whether SAP runs on this host"

ryandillinfelton: Added

PR #2, opened 2020-07-16
Remove $id

Glutexo:
The $id key is not compatible with OpenAPI. Remove it.

In https://github.com/RedHatInsights/insights-host-inventory/pull/682, an exact copy of this file starts to be used in the OpenAPI specification. Syncing.

Hopefully the _$id_ won’t be necessary to use _jsonschema_ or other libraries directly.


iphands:
Dont know the details here... but it seems sane to me :D

PR #3, opened 2020-07-24
Add sap_sids fact

CodeHeeler:


jharting:
@CodeHeeler please add a description of the new field and ideally also an example value. Are there any constraints that should be validated on array items besides them being strings?

Glutexo:
May I ask for a source for this new field? Where was it added to the payload?

beav:
It was added in puptoo in https://github.com/RedHatInsights/insights-puptoo/blob/master/src/puptoo/process.py#L125-L127, along with the `sap_system` field.

re: constraints, we are working on gathering the max length and examples. I don't expect there to be more than a handful of strings in the array, with each string being less than 100-200 char.

Glutexo:
Thanks! Would you add that to the specification together with a description?

CodeHeeler:
> Thanks! Would you add that to the specification together with a description?

Added description, example, maxLength, and pattern/regex based on https://help.sap.com/viewer/2c1988d620e04368aa4103bf26f17727/2.0.05/en-US/7672838aa9c44c7ab9d10188531669f3.html

beav (approved):

jharting (changes requested):
@CodeHeeler @beav are you sure about the format? Here are a few examples of what puptoo is sending: (HDB00, HDB10, SCS02, SMDA98)

cc @SteveHNH

beav:
@jharting good catch. it looks like there are two different attributes we are conflating.

There is the `local_instances`, which are the instances that are running on the server. These are not the SIDs.

We probably need to rename this field to `sap_sids`. The validation should stay the same.

I think in Puptoo, it should do something like `profile["sap_sids"] = [sap.sid(instance) for instance in local_instances]` (untested pseudocode).

beav:
Here is what I have with the proposed puptoo change:

```
'sap_system': True,
'sap_sids': ['RHE']
```

Most systems will only have one SID, but its possible to have multiple.

beav:
https://github.com/RedHatInsights/insights-puptoo/pull/67

It's OK to merge the puptoo PR before any inventory change. Extra fields are ignored by inventory svc.

beav:
related: https://github.com/RedHatInsights/insights-host-inventory/pull/709

schemas/system_profile/v1.yaml
```diff
@@ -239,3 +239,11 @@ $defs:
       sap_system:
         type: boolean
         description: Indicates if SAP is installed on the system
+      sap_sids:
+        type: array
+        items:
+          description: The SAP system ID (SID)
+          type: string
+          example: "H2O"
```
Glutexo, line 247: Is this an example of a real value? 😃

schemas/system_profile/v1.yaml
```diff
@@ -239,3 +239,11 @@ $defs:
...
+          maxLength: 3
```
Glutexo, line 248: This is slightly redundant, because the _pattern_ already demands a string of an exact length of 3 characters. For consistency we could even add `minLength: 3`.

Glutexo (approved):
I can approve like this, but I’d consider adding the _minLength_ constraint. Also please take a look at my last [comment](https://github.com/RedHatInsights/insights-host-inventory/pull/709#issuecomment-690938713) for https://github.com/RedHatInsights/insights-host-inventory/pull/709. We need the regex validation in Marshmallow too.

schemas/system_profile/v1.yaml
CodeHeeler: heh, I had no examples of real values at that time, but since it only demands the first character be a letter, and the subsequent two can be digits or letters, it's actually a possible real value. :)

schemas/system_profile/v1.yaml
CodeHeeler: The pattern will handle everything, but it seemed we generally added maxLength in the schema, so I put it there for ease of folks glancing through the code to instantly know the space these are taking up rather than having to parse the regex to figure it out.  I don't see why we should also add the minLength, but I am happy to do so if you'd prefer.

CodeHeeler:
> I can approve like this, but I’d consider adding the _minLength_ constraint. Also please take a look at my last [comment](https://github.com/RedHatInsights/insights-host-inventory/pull/709#issuecomment-690938713) for [RedHatInsights/insights-host-inventory#709](https://github.com/RedHatInsights/insights-host-inventory/pull/709). We need the regex validation in Marshmallow too.

What is Marshmallow?  Where is the repo for that?  Was this comment meant for @beav or do you know which team owns Marshmallow so that I have a contact on this work?  Thanks!

jharting (approved):

PR #5, opened 2020-07-27
RHCLOUD-7977 set up CI for inventory-schemas

jharting:


.spectral.yaml
```diff
@@ -0,0 +1,68 @@
...
+    given: $..properties[?(@.type == 'string' || @.type == 'integer' || @.type == 'number' || @.type == 'boolean')]
+    severity: error
+    then:
+      field: description
+      function: truthy
+
+  property-example:
+    description: Every property should define an example
+    message: "{{path}} does not define an example"
+    given: $..properties[?(@.type == 'string' || @.type == 'integer' || @.type == 'number')]
```
Glutexo, line 25: It may make some sense to require this for collection items and nested object properties. It’s probably still ok to have a mandatory example only for the root item – if it’s good, it contains all the nested fields already.

.spectral.yaml
```diff
@@ -0,0 +1,68 @@
...
+    severity: error
+    then:
+      field: example
+      function: truthy
+
+# TODO: remove these exceptions as we clean up the schema and clarify  in RHCLOUD-8147
```
Glutexo, line 31: This is an awesome way to have a list of entries that don’t comply! 🙇🏻‍♂️

I see that description of many properties is totally not useful. I mean like “mount_point: mount point”. Maybe we can add the same rules for those fields? Even though they would be technically useless, they’d have an informational purpose.

schemas/system_profile/v1.yaml
```diff
@@ -1,5 +1,6 @@
 ---
 $id: system_profile.spec.yaml
+$schema: https://json-schema.org/draft/2019-09/schema#
```
Glutexo, line 3: I wonder whether it’s possible to pass the schema type as a CLI argument. I didn’t found a way though.

Jenkinsfile
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env groovy
+
+/*
+ * Not using https://github.com/marketplace/actions/spectral-linting
+ * because of https://github.com/stoplightio/spectral-action/issues/200.
+ */
+
+node('nodejs') {
+
+    env.NODEJS_HOME = "${tool 'node-10'}"
```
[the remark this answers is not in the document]
jharting: https://plugins.jenkins.io/nodejs/

Jenkinsfile
```diff
@@ -0,0 +1,22 @@
...
+    env.PATH="${env.NODEJS_HOME}/bin:${env.PATH}"
+
+    checkout scm
+
+    stage('Install Spectral') {
+        sh 'npm i -g @stoplight/spectral'
```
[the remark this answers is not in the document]
jharting: https://plugins.jenkins.io/nodejs/

schemas/system_profile/v1.yaml
jharting: Neither did I. Having it contained withing the document is better anyway IMO.

.spectral.yaml
jharting: Let's add a rule in a follow-up PR that checks if a description at least resembles a sentence.

.spectral.yaml
```diff
@@ -0,0 +1,68 @@
...
+    severity: error
+    then:
+      function: pattern
+      functionOptions:
+        match: '^[a-z]+[a-z0-9_]*[a-z0-9]+$'
+
+  property-description:
+    description: Every property should define a description
+    message: "{{path}} does not define description"
+    given: $..properties[?(@.type == 'string' || @.type == 'integer' || @.type == 'number' || @.type == 'boolean')]
```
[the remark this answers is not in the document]
jharting: Updated the query to match non-array properties and items and switched to negative condition

.spectral.yaml
```diff
@@ -0,0 +1,75 @@
...
+    severity: error
+    then:
+      function: pattern
+      functionOptions:
+        match: '^[a-z]+[a-z0-9_]*[a-z0-9]+$'
+
+  property-description:
+    description: Every non-array type should define a description
+    message: "{{path}} does not define description"
+    given: $..[?(@property=='items' || (@parentProperty=='properties' && @.type != 'array'))]
```
Glutexo, line 16: I like that when the _items_ contain a _$ref_, the description is taken from the referenced definition. 👍🏻

If the expression is regrouped, it will also catch arrays items that are of type _array_ themselves. `((@property=='items' || @parentProperty=='properties') && @.type != 'array')` It is not used anywhere, but it would be consistent with the expression in the _property-example_ rule below.

.spectral.yaml
```diff
@@ -0,0 +1,75 @@
...
+    severity: error
+    then:
+      field: description
+      function: truthy
+
+  property-example:
+    description: Every string or numeric property should define an example
+    message: "{{path}} does not define an example"
+    given: $..[?((@property=='items' || @parentProperty=='properties') && (@.type == 'string' || @.type == 'integer' || @.type == 'number'))]
```
Glutexo, line 25: - Not for `@.type == 'object'`? We agreed that we want to disallow those entirely, but currently there is one (_$defs/DiskDevice/options_) and it even has an example. 🤷🏻‍♂️

- This rule contains a positive list of matched types. The other rule, _property-description_, on the other hand uses a negative list `@.type != 'array'`. Using such one here as well `@.type != 'array' && @.type != 'boolean' && @.type != 'object'` would be in my opinion more readable. And it would raise a question, why exclude the _object_ type. 😃

.spectral.yaml
```diff
@@ -0,0 +1,75 @@
...
+    description: Every non-array type should define a description
+    message: "{{path}} does not define description"
+    given: $..[?(@property=='items' || (@parentProperty=='properties' && @.type != 'array'))]
+    severity: error
+    then:
+      field: description
+      function: truthy
+
+  property-example:
+    description: Every string or numeric property should define an example
```
Glutexo, line 23: If the type list would be converted to be negative, this would need to be changed too.

.spectral.yaml
jharting: As long as object properties have examples there is no need for the object to also have an example. It's allowed but not required. Inverted the expression as requested.

.spectral.yaml
jharting: ok

.spectral.yaml
```diff
@@ -0,0 +1,75 @@
...
+    given: $..[?((@property=='items' || @parentProperty=='properties') && @.type != 'array')]
+    severity: error
+    then:
+      field: description
+      function: truthy
+
+  property-example:
+    description: Scalar types (except for boolean) should define an example
+    message: "{{path}} does not define an example"
+    given: $..[?((@property=='items' || @parentProperty=='properties') && @.type != 'array' && @.type != 'object' && @.type != 'boolean' && @.type != null)]
```
Glutexo, line 25: Is `@.type != null` matching missing type or just an explicit null?

PR #7, opened 2020-07-30
Remove unnecessary stuff

Glutexo:
- The titles don’t add any value. They just repeat what the names say. Removed.
- Definitions with schemas are objects by default. Use object only for arbitrary dictionaries. (Those are going to be removed anyway.)
- Version is not necessary. For now there is going to be only one canonical schema.

No JIRA ticket. Slightly related to [RHCLOUD-8147](https://projects.engineering.redhat.com/browse/RHCLOUD-8147).

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -116,7 +116,6 @@ $defs:
         maxLength: 18
   SystemProfile:
     description: Representation of the system profile fields
-    type: object
     properties:
       number_of_cpus:
         type: integer
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -1,7 +1,6 @@
 ---
 $id: system_profile.spec.yaml
 $schema: https://json-schema.org/draft/2019-09/schema#
-$version: 1
 $defs:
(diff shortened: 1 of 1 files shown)


jharting:
@Glutexo I know it's tempting to start to clean up the spec but let's first make sure to get the spec on par with what's currently used in inventory and only then do any further changes.

Glutexo:
Do these changes by any means conflict with the actual definitions?

jharting:
Reasoning about whether we succeeded with getting this schema on part with what's currently used in inventory is going to be much simpler if they don't diverge in the meantime.

Glutexo:
Rebased on current master.

Glutexo:
May I ask @shdunning for a review?

schemas/system_profile/v1.yaml
```diff
@@ -1,10 +1,8 @@
 ---
 $id: system_profile.spec.yaml
 $schema: https://json-schema.org/draft/2019-09/schema#
-$version: 1
```
shdunning, line 4: not sure if it's related, but figuring out how to "migrate" from one schema to the next is going to be an upcoming topic as part of the LTB effort.

PR #9, opened 2020-08-06
Add os_release_major, os_release_minor

iphands:
I would like to let reporters start splitting out the os_release string into two
(for now) ints. We can talk about deprecating the old os_release String once we
have report quorum/adoption.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -161,6 +161,12 @@ $defs:
           type: string
           maxLength: 30
         type: array
+      os_release_major:
+        type: int
+        maxLength: 4
+      os_release_minor:
+        type: int
+        maxLength: 4
       os_release:
         type: string
         maxLength: 100


schemas/system_profile/v1.yaml
```diff
@@ -161,6 +161,12 @@ $defs:
           type: string
           maxLength: 30
         type: array
+      os_release_major:
+        type: int
+        maxLength: 4
+      os_release_minor:
+        type: int
+        maxLength: 4
       os_release:
```
ShimShtein, line 170: What info should live under the `os_release` field then?

iphands: I imagine it can stick around with he NN.NN string format for backwards compatibility.
We should strive not to break things, or break them veerrryyy slooowwwly.

Thoughts about that though are appreciated.

ShimShtein: It looks like we want to store not only RHEL hosts, so it would be awesome if we had the OS name and OS flavor (server/desktop e.t.c) stored.

ShimShtein: Maybe it would be even better to add those fields as a child to a general `operating_system` tag:

``` json
"operating_system" : {
  "major": 7,
  "minor": 6,
  "name": "RHEL",
  "modification": "server"
}
```

iphands: Ohhh I think the more than this or #26 
@CodeHeeler @beav thoughts?

iphands: Shim, name scares me though.
Isnt the name for desktop really Red Hat Enterprise Linux for Workstations? or RHELW.
I know Suse has SLES and SLED.

I want to make sure that the name, mod are enums
Perhaps we say:
```
vendor:
name:
modification:
```
??

That way as an app I can filter by vendor and be assured its some flavor of RHEL?

iphands: Im curious if we prefer something like:
vendor enum: [ "Red Hat", "CentOS", "Canonnical", "Debian"]
VS
flavor enum: ["RHEL", "SLES", "Ubuntu", "Fedora", "Debian"]

beav: I'm :+1: to flavor instead of vendor. If it's something we will be displaying to end users like via the inventory app frontend UI, it has to be "Red Hat Enterprise Linux" spelled out.

I would stick with just "Red Hat Enterprise Linux" and "Fedora" for now since we know how to detect those.

I am not sure how to describe server/workstation/hpc/etc in a way that isn't specific to rhel. Would it work if we left it out for now, and just did flavor?

Also I would do major/minor as strings in case there are minor versions with leading zeroes (5.01)

iphands:
Closing in favor of #27

PR #10, opened 2020-08-06
adding simple testing tools

iphands:
Just adding some k.i.s.s. tool for testing schema changes against a single test JSON file


tools/simple-test/tester.py
```diff
@@ -0,0 +1,13 @@
+import yaml
+import json
+import sys
+from jsonschema import validate
+
+schema_path = sys.argv[1]
+sample_path = sys.argv[2]
+
+with open(sample_path, 'r') as sample_data:
+    with open(schema_path, 'r') as schema_data:
```
Glutexo: - _r_ is the default mode. ([reference](https://docs.python.org/ja/3.6/library/functions.html#open))
- _with_ can have multiple clauses and save indentation.

```suggestion
with open(sample_path) as sample_data, open(schema_path) as schema_data:
```

tools/simple-test/tester.py
```diff
@@ -0,0 +1,13 @@
...
+        schema_dict = yaml.safe_load(schema_data)
+        sample_dict = json.load(sample_data)
+        validate(instance=sample_dict, schema=schema_dict)
```
Glutexo: This actually doesn’t work. The reason is that the schema specification file only contains definitions. To validate a document, you need a schema that uses the SystemProfile definition. This can be easily achieved by adding `"$ref": "#/$defs/SystemProfile"` to the specification root. This needs not to be added to the file, sticking it up there in the in-memory representation is fine.

```suggestion
    with open(spec_path, 'r') as spec_data:
        spec_dict = yaml_safe_load(spec_data)
        schema_dict = {**spec_dict, "$ref": "#/$defs/SystemProfile"}
        sample_dict = json_load(sample_data)
        jsonschema_validate(instance=sample_dict, schema=schema_dict)
```

Glutexo: It’s actually not necessary to enclose the _validate_ call in the _open_ context. The files are already loaded and can be closed by that time.

```suggestion
with open(sample_path, 'r') as sample_data:
    with open(schema_path, 'r') as schema_data:
        schema_dict = yaml.safe_load(schema_data)
        sample_dict = json.load(sample_data)
validate(instance=sample_dict, schema=schema_dict)
```

PR #12, opened 2020-08-12
Validate against Draft4

Glutexo:
The Host Inventory uses JSONSchema Draft 4 for validation. [[Reference](https://github.com/zalando/connexion/blob/3fecd3e7351bfff330c5421551d8e26f88cbfc28/connexion/decorators/validation.py#L109)] It is the default of the Connexion framework used by the app. Use the same validator here to ensure compatibility.

### DIFF
--- a/tools/simple-test/tester.py
+++ b/tools/simple-test/tester.py
@@ -2,6 +2,7 @@ from sys import argv
 
 from json import load as json_load
 from jsonschema import validate as jsonschema_validate
+from jsonschema.validators import Draft4Validator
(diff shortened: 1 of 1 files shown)


Glutexo:
We don’t even need to change this to Draft7. Once the version in the document matches with the validator used, it is picked automatically. [[Reference](https://github.com/Julian/jsonschema/blob/2f734a7ce1395ac4ff2d394583fd46a2e8833a9b/jsonschema/validators.py#L904)]

Glutexo:
Replaced by #15.

PR #13, opened 2020-08-12
Make tester executable

Glutexo:
Turned the JSON Schema tester to a proper standalone script:

- Used argparse to parse command line arguments.
- Added `__name__ == "__main__"` conditional to make it importable.
- Chmodded u+x to make it executable.

_run.sh_ still works without change.

Possible usage:

```bash
$ . tools/simple-test/venv/activate
$ tools/simple-test/tester.py ../../schemas/system_profile/v1.yaml ../../scratch/sample.json
```

### DIFF
--- a/tools/simple-test/tester.py
+++ b/tools/simple-test/tester.py
@@ -1,15 +1,28 @@
-from sys import argv
+#!/usr/bin/env python
+
+from argparse import ArgumentParser
 
 from json import load as json_load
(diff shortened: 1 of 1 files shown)


Glutexo:
Python executables are nicely executable by PyCharm run configurations. Bash scripts are too, but why do things in bash when the IDE can do them with all its awareness.

PR #14, opened 2020-08-12
Disk device options

Glutexo:
Added a constraint for _/disk_devices/options_. In Marshmallow we require all keys to be non-empty strings, including those of the nested objects. The new constraint does exactly this.

Examples of invalid payloads:

```json
  "disk_devices": [
    {
      "type": "ext3",
      "label": "home drive",
      "device": "/dev/sdb1",
      "options": {
        "ro": true,
        "uid": "0",
        "": "thou shall not pass"
      },
      "mount_point": "/home"
    }
  ],
```

```json
  "disk_devices": [
    {
      "type": "ext3",
      "label": "home drive",
      "device": "/dev/sdb1",
      "options": {
        "ro": true,
        "uid": "0",
        "nested": {
            "this": "is fine",
            "": "is not"
        }
      },
      "mount_point": "/home"
    }
  ],
```

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -2,6 +2,11 @@
 $id: system_profile.spec.yaml
 $schema: https://json-schema.org/draft/2019-09/schema#
 $defs:
+  NestedObject:
+    propertyNames:
+      minLength: 1
+    additionalProperties:
+      "$ref": "#/$defs/NestedObject"
   DiskDevice:
     description: Representation of one mounted device
     properties:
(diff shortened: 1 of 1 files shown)


Glutexo:
Conquered the linter. There really was a missing description. Fixed.

Glutexo:
In the end, the recursion introduced in this pull request caused a problem. _prance_ is unfortunately unable to dereference the system profile schema if it contains recursion like this. The reason is that first the internal references in a referenced file are converted to file references and are no longer considered internal.

Glutexo:
It actually makes perfect sense. It is not possible to keep the local references, because the references schemas are no longer locally available. It would be possible to insert them somewhere locally and then convert the local reference path. Prance however does not do this.

Created an issue for _prance_ to ask the author what does he think about this. https://github.com/jfinkhaeuser/prance/issues/77 We can’t however expect this to be resolved by the library. My suggestion is to ditch prance and dereference manually.

PR #16, opened 2020-08-21
Make objects explicit again

Glutexo:
I was wrong in #7 claiming that there is no need to explicitly define object types.

If there is no type defined, any type is valid. In case of objects, defining properties and stuff is not enough. Those are used only if the respective document is an object, but non-objects are still valid. Added explicit type definitions to all objects.

This also helps Connexion’s _coerce_type_ function to correctly process values of an object. The function is going to be used in the Host Inventory.

### Minimal working example ###

```python
from jsonschema import validate
schema = {"properties": {"property": {"minLength": 1}}}
validate({"property": ""}, schema)  # invalid
validate([""], schema)  # valid
```

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -3,6 +3,7 @@ $id: system_profile.spec.yaml
 $schema: http://json-schema.org/draft-07/schema#
 $defs:
   NestedObject:
+    type: object
     description: An arbitrary object that doesn’t allow empty string keys.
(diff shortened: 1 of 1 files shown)


Glutexo:
It should have been an objection. 😃 The type definitions should have been there! Fixed, pushed. Nice catch, thank you!

This still needs one more review. @jharting, @shdunning, @tahmidefaz?

jharting (approved):

PR #17, opened 2020-09-21
add tuned_profile to facts

CodeHeeler:


beav:
Is there a list of what these can be? like "desktop", "server", "toaster", etc

CodeHeeler:
> Is there a list of what these can be? like "desktop", "server", "toaster", etc
> 
> edit: it looks like you can create your own name, and the name is based on the directory name of the tuned profile. I would bump this from 50 char to 256 char max length (255 = max chars in a filename)

Exactly, it can be anything, see this image below for some more examples which is why I went with the length I did.  Happy to bump up the char limit.

image

schemas/system_profile/v1.yaml
```diff
@@ -247,3 +247,8 @@ $defs:
           example: "H2O"
           maxLength: 3
           pattern: '^[A-Z][A-Z0-9]{2}$'
+      tuned_profile:
+        type: string
+        maxLength: 256
+        description: Current profile resulting from command tuned-adm active
+        example: 'desktop'
```
beav, line 254: just to show you can have a custom string, can you add an example like "my-super-cool-custom-profile"?

CodeHeeler: So something like this? (Using how to add multiple examples here: https://swagger.io/docs/specification/adding-examples/)

```type: string
maxLength: 256
description: Current profile resulting from command tuned-adm active
examples:
    common:
        summary: A typical tuned profile name
        value: 'desktop'
    custom:
        summary: Any custom name will do
        value: 'my-super-cool-custom-profile'
```

beav: ah! i was not aware the examples had a particular syntax :) ya that works

beav:
it looks like something is unhappy with the yaml. Can you try:

`npm i -g @stoplight/spectral; spectral lint schemas/system_profile/v1.yaml` ?

CodeHeeler:
> it looks like something is unhappy with the yaml. Can you try:
> 
> `npm i -g @stoplight/spectral; spectral lint schemas/system_profile/v1.yaml` ?

Looks like maybe the linter is set to want `example` so my change to include `examples` as the swagger docs instructed is breaking it.  Not sure whether you'd prefer to go back to one example or find a way to tell the linter this is ok too, or if there's something else I can try.

```
(inventory-schemas) [dawalker@localhost inventory-schemas]$ spectral lint schemas/system_profile/v1.yaml
JSON Schema detected
JSON Schema (loose) detected
JSON Schema Draft 7 detected/home/dawalker/devel/inventory-schemas/schemas/system_profile/v1.yaml
 250:21  error  property-example  #/$defs/SystemProfile/properties/tuned_profile does not define an example✖ 1 problem (1 error, 0 warnings, 0 infos, 0 hints)
```

beav:
> > it looks like something is unhappy with the yaml. Can you try:
> > `npm i -g @stoplight/spectral; spectral lint schemas/system_profile/v1.yaml` ?
> 
> Looks like maybe the linter is set to want `example` so my change to include `examples` as the swagger docs instructed is breaking it. Not sure whether you'd prefer to go back to one example or find a way to tell the linter this is ok too, or if there's something else I can try.
> 
> ```
> (inventory-schemas) [dawalker@localhost inventory-schemas]$ spectral lint schemas/system_profile/v1.yaml
> JSON Schema detected
> JSON Schema (loose) detected
> JSON Schema Draft 7 detected
> 
> /home/dawalker/devel/inventory-schemas/schemas/system_profile/v1.yaml
>  250:21  error  property-example  #/$defs/SystemProfile/properties/tuned_profile does not define an example
> 
> ✖ 1 problem (1 error, 0 warnings, 0 infos, 0 hints)
> ```

ya just go back to the old way. My prior suggestion was just so people realize you can have user-defined strings but if the linter is being annoying, its better to make the linter happy.

beav:
this change looks good to me. The corresponding change in the current schema is in https://github.com/RedHatInsights/insights-host-inventory/pull/719.

beav (approved):

beav:
@Glutexo @jharting if the changes above look ok and you don't have further q's, can you merge?

thanks!

tahmidefaz (approved):

PR #18, opened 2020-09-25
support subscription-manager release lock

MichaelMraka:
added rhsm_releasever to support for subscription-manager release lock


shdunning:
@MichaelMraka need to add `description` and `example` to your new field.

schemas/system_profile/v1.yaml
```diff
@@ -247,3 +247,6 @@ $defs:
           example: "H2O"
           maxLength: 3
           pattern: '^[A-Z][A-Z0-9]{2}$'
+      rhsm_releasever:
```
shdunning, line 250: Need to add `description` and `example`.

```
JSON Schema detected
JSON Schema (loose) detected
JSON Schema Draft 7 detected
 
/tmp/workspace/inventory-schemas_PR-18/schemas/system_profile/v1.yaml
250:23 error property-description #/$defs/SystemProfile/properties/rhsm_releasever does not define description
250:23 error property-example #/$defs/SystemProfile/properties/rhsm_releasever does not define an example
 
2 problems (2 errors, 0 warnings, 0 infos, 0 hints) 
script returned exit code 1
```

shdunning (changes requested):

schemas/system_profile/v1.yaml
MichaelMraka: comment and example fixed

schemas/system_profile/v1.yaml
```diff
@@ -247,3 +247,8 @@ $defs:
...
+        description: Sytem release version set by subscription-manager
```
jharting, line 251: Sytem

schemas/system_profile/v1.yaml
```diff
@@ -247,3 +247,8 @@ $defs:
...
+        type: string
+        example: "8.2"
```
jharting, line 253: Does subscription-manager always follow the `major.minor` format? If the output is well-defined I wonder if we can specify it here with a regex

jharting (changes requested):

schemas/system_profile/v1.yaml
MichaelMraka: No, it can be "8", "8.0", "8.1", ... or empty (""). Or it might not be even set (IMHO majority of the cases).

schemas/system_profile/v1.yaml
MichaelMraka: fixed typo

MichaelMraka:
Rebased to current master and squashed fixes into a single commit.

iphands:
@MichaelMraka  I *think* this is unneeded after we merged https://github.com/RedHatInsights/inventory-schemas/pull/27
Closing please do open if you think this is incorrect.

PR #19, opened 2020-09-28
Adding more lint - RHCLOUD-8254

tahmidefaz:


tahmidefaz:
Lmk, if you think we should add any more lint, atm!

kruai (approved):
LGTM!

schemas/system_profile/v1.yaml
```diff
@@ -91,32 +91,34 @@ $defs:
         items:
           type: string
           format: ipv4
+          maxLength: 15
```
jharting, line 94: This is potentially a breaking change. (Even though it looks unlikely we saw reporters sending unexpected stuff in the past). It would be better to split this into linter PR (where linter violations caused by existing definitions are suppressed by the exclusion list) and a follow-up PR where these are addressed (after verifying that the change does not break existing reporters).

tahmidefaz: :lion: :tiger: :bear: :open_mouth:

thearifismail (approved):

.spectral.yaml
```diff
@@ -28,6 +28,34 @@ rules:
...
+
+  property-field-sentence:
+    description: Description of a field should be a sentence
+    message: "{{path}} is not a sentence"
+    given: $..[?(@property=='description')]
+    severity: error
+    then:
+      function: pattern
+      functionOptions:
+        match: '^[A-Z]+\s*\S+(?:\s+\S+){2,}\s*$'
```
rafaelcassau, line 48:
```suggestion
        match: '^[A-Z]*\s*\S+(?:\s+\S+){2,}\s*$'
```
In this case the expression `A valid expression` will fail because of `^[A-Z]+` because it requires one more character, to work properly you can replace `+` by `*`, in this case `+` means one or more while `*` means zero or more.

rafaelcassau (approved):
LGTM

tahmidefaz:
@ryandillinfelton added `package.json` like you suggested.

ryandillinfelton (changes requested):
Just another change to the dang regex. But looks/works great otherwise

.spectral.yaml
```diff
@@ -28,6 +28,34 @@ rules:
...
+
+  property-field-sentence:
+    description: Description of a field should be a sentence
+    message: "{{path}} is not a sentence"
+    given: $..[?(@property=='description')]
+    severity: error
+    then:
+      function: pattern
+      functionOptions:
+        match: '^[A-Z]*\s*\S+(?:\s+\S+){2,}\s*$'
```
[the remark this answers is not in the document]
rafaelcassau: it's a good improvement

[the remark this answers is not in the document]
tahmidefaz: Just updated the regex like you suggested. Thanks a lot! :+1:

ryandillinfelton (approved):
LGTM :rocket:

PR #21, opened 2020-10-01
Fix yum_repos/base_url field name

Glutexo:
By a mistake, the _yum_repos/base_url_ key had a wrong name. @rodrigonull reported this. Unfortunately this even got to the Host Inventory codebase, causing this value not to be stored at all. 😞 A fix for this in the Host Inventory is a part of https://github.com/RedHatInsights/insights-host-inventory/pull/715. Fixing here too for consistency.


kruai (approved):

tahmidefaz (approved):

tahmidefaz:
Merging this because this is how it works in prod as of today and nothing is on :fire: .

PR #22, opened 2020-10-05
Add query generation script

tahmidefaz:
The goal of this script is to make it easier for us to open query schema MRs with app-sre. This script will generate sql query schema YAML that you can copy over and make a MR for app-sre.

Features:
- [x] Max length query for higher-level strings (not the ones nested in an object or array).
- [x] Pattern matching for higher-level strings.
- [x] Max length & pattern queries for lower-level strings.
- [x] Add docs for script usage

Usage:
`python3 query_gen.py <field_name|path> <filed_type (str|arr|nested-arr|nested-str)> <operation (max-length|pattern)> <value>`

The queries, when submitted to app-sre, will produce results such as:
image

~~Let me know if you guys have any idea how to query strings in nested JSON object or array, I am still researching it! :crossed_fingers:~~ 

**Update:** added support for querying arrays and nested strings


thearifismail:
@tahmidefaz I am trying to understand the goal mentioned in beginning.  Is the only goal to produce a query that AppSRE can run against an RDS DB and give us the result back?  How does user get the result back?  Review the Jenkins log?

About the module name `queries`, at the moment it has `SelectQuery` only; what other queries are expected to be added in future?  If the answer is None; then why not name the module.

Need to document, what is expected from the user to know when using `queries or SelectQuery`.  e.g. have a running DB like HBI.  May be a working example of query

tahmidefaz:
> @tahmidefaz I am trying to understand the goal mentioned in beginning. Is the only goal to produce a query that AppSRE can run against an RDS DB and give us the result back? How does user get the result back? Review the Jenkins log?

The user needs to look at the logs produced by the pod that ran on OSDv4 upon the submission of the yaml file.

> 
> About the module name `queries`, at the moment it has `SelectQuery` only; what other queries are expected to be added in future? If the answer is None; then why not name the module.

Named it `SelectQueries` for accuracy, not sure what you meant by naming the module. Can you please elaborate? 
> 
> Need to document, what is expected from the user to know when using `queries or SelectQuery`. e.g. have a running DB like HBI. May be a working example of query

Just added some documentation, let me know if this is not enough.


@thearifismail please check my inline replies.

thearifismail:
Need a sentence about why a user submits an MR to`app-sre`.  Answer is that app owners do not have permissions to run queries against RDS databases deployed by app-sre.

Needs:

1. if some service other than HBI wants to use, how SelectQuery should be modified?
2. `TODO` section listing what more can be done

tahmidefaz:
> Needs:
> 
> 1. if some service other than HBI wants to use, how SelectQuery should be modified?
> 2. `TODO` section listing what more can be done

@thearifismail after our conversation this morning, I thought about it a little. I honestly don't think other services can easily make use of this script. It won't be as simple as modifying the `SelectQueries` since all our required arguments are baked into the code. They might be better off writing their own.

To answer no. 2, I am unsure how else we can expand it as these queries pretty much cover all that we need, imo :slightly_smiling_face:

thearifismail:
> @thearifismail after our conversation this morning, I thought about it a little. I honestly don't think other services can easily make use of this script. It won't be as simple as modifying the SelectQueries since all our required arguments are baked into the code. They might be better off writing their own.

Agree.  Before seeing your response to my comment,  I realized that the last time we realized that App-SRE help would be needed was for deleting an `orphan replication_slot`, which would not be covered by thiw PR. That's when I realized how quickly its complexities can grown.  To make the story, this is good for now

thearifismail (approved):

kruai (approved):

PR #25, opened 2020-10-13
Add additional sap facts

CodeHeeler:
Puptoo PR here: RedHatInsights/insights-puptoo#72


CodeHeeler:
Puptoo PR here: https://github.com/RedHatInsights/insights-puptoo/pull/72

kruai (approved):

schemas/system_profile/v1.yaml
```diff
@@ -251,6 +251,18 @@ $defs:
           example: "H2O"
           maxLength: 3
           pattern: '^[A-Z][A-Z0-9]{2}$'
+      sap_instance_number:
+        type: string
+        description: The instance number of the SAP HANA system
+        example: '42'
+        maxLength: 2
+        pattern: '(?!99)(?!98)^[0-9]{2}$'
+      sap_version_info:
```
gmcculloug, line 260: @CodeHeeler Is there a reason to include "_info" in this key name?  Is it following some existing pattern?

This schema file current has: `bios_version`, `os_kernel_version`, `insights_client_version`, and `insights_egg_version`.

Could this key be reduced to `sap_version`?

jeromemarc: Let's keep it consistent with other fact names. Let's go with `sap_version`.

CodeHeeler: Was following @jeromemarc's lead and that was how it was initially described.  Just pushed up the change to the new shorter name now.

iphands:
@CodeHeeler would we rather make this an integer type?
Its probably much faster to parse if it is.

Is FE, A2. D8 ever supported?

CodeHeeler:
> @CodeHeeler would we rather make this an integer type?
> Its probably much faster to parse if it is.

Sure, I'll update it to integer.

> Is FE, A2. D8 ever supported?

Not sure what this means? ^

schemas/system_profile/v1.yaml
```diff
@@ -251,6 +251,18 @@ $defs:
           example: "H2O"
           maxLength: 3
           pattern: '^[A-Z][A-Z0-9]{2}$'
+      sap_instance_number:
+        type: integer
+        description: The instance number of the SAP HANA system
+        example: '42'
+        maxLength: 2
+        pattern: '(?!99)(?!98)^[0-9]{2}$'
```
tahmidefaz, line 259: since this is an integer now, I think using `maximum` and `minimum` would be ideal here instead of `maxLength`. You also won't need the `pattern` anymore.

tahmidefaz (changes requested):

schemas/system_profile/v1.yaml
CodeHeeler: Actually now that I think about it and look back at the spec, what about 00, 01, 02, etc?  Comparisons won't change, but is there any use case for wanting that leading zero included instead of using only an integer?

CodeHeeler: See this source: https://help.sap.com/viewer/2c1988d620e04368aa4103bf26f17727/2.0.05/en-US/ae305ab18bad42efa7b067e7cc82b846.html

iphands: WAIT, if 00 - 09 are valid, has to be String again right? (thanks Dana :D)

ryandillinfelton (approved):

tahmidefaz (approved):

schemas/system_profile/v1.yaml
```diff
@@ -251,6 +251,18 @@ $defs:
           example: "H2O"
           maxLength: 3
           pattern: '^[A-Z][A-Z0-9]{2}$'
+      sap_instance_number:
+        type: string
+        description: The instance number of the SAP HANA system
+        example: '42'
```
gmcculloug, line 257: @CodeHeeler For clarify maybe add the constraint to the description:
```yaml
description: The instance number of the SAP HANA system (must be a two-digit number between 00 and 97)
```

Also, the example value could be extended to show a leading-zero value in addition to `03 or 42`.

CodeHeeler: Done. :)

PR #26, opened 2020-10-21
Add os_vendor enum

iphands:
Applications (inventory at least) want to be able to filter out only RHEL like systems.
We don't want to necessarily differentiate between os types (ie RHEL vs RH Desktop Linux etc)...
Or not at the moment at least.

We could still add in the future os_type, os_release, os_flavor etc (or something like those).

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -165,6 +165,10 @@ $defs:
           type: string
           maxLength: 30
         type: array
+      os_vendor:
+        type: string
+        enum: ["Red Hat", "Cannonical", "Oracle", "CentOS", "Debian"]
+        description: The vendor of the OS/distro
       os_release:
         type: string
         maxLength: 100
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -167,6 +167,7 @@ $defs:
         type: array
       os_vendor:
         type: string
+        example: "Red Hat"
         enum: ["Red Hat", "Cannonical", "Oracle", "CentOS", "Debian"]
         description: The vendor of the OS/distro
       os_release:


schemas/system_profile/v1.yaml
```diff
@@ -165,6 +165,10 @@ $defs:
           type: string
           maxLength: 30
         type: array
+      os_vendor:
+        type: string
+        enum: ["Red Hat", "Cannonical", "Oracle", "CentOS", "Debian"]
```
CodeHeeler, line 170: Are these the only possible vendors?

CodeHeeler (approved):
Other than double checking that all os's are exclusively from one of the listed vendors, this LGTM.

schemas/system_profile/v1.yaml
beav: I don't think this is exposed as a list of possible strings right now by insights-core.

I think the enum is OK but we may want to just stick with "Red Hat" as the only value for now. It can be filled in if a system either has values set in the redhat_release combiner, or of it has the correct installed product (not sure which is best).

Later we may want to add HyperV and a few other hypervisors.

sthirugn, line 170: s/Cannonical/Canonical?

iphands: I dont know the set for now... FWIW I think now, today CentOS and such are being shoved in from Sat.
BK asked about whether we wanted it or not.

Limiting it to "Red Hat" would solve that in that it would be an explicit decision to let in others if we want.
IOW we could start throwing away all non RHEL until we want them for some reason.

iphands: Fixed in #27

iphands:
Closing in favor of #27

PR #27, opened 2020-10-23
Combining #9 and #26 splitting os release and adding os vendor flavor etc

iphands:
Do we want to do something more like this? instead of #26 and #9 

@jharting do you think flatter is better for searching
cc @gmcculloug on the searching thing

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -10,6 +10,7 @@ $defs:
     additionalProperties:
       "$ref": "#/$defs/NestedObject"
   OperatingSystem:
+    description: Object for OS details
     type: object
     properties:
       major:
@@ -24,8 +25,9 @@ $defs:
         maxLength: 4
       name:
         description: Name of the distro/os
-        example: "RHEL"
+        example: RHEL
         type: string
+        maxLength: 4
         enum: ["RHEL"]
   DiskDevice:
     description: Representation of one mounted device
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -27,6 +27,7 @@ $defs:
         description: Name of the distro/os
         example: RHEL
         type: string
+        maxLength: 4
         enum: ["RHEL"]
   DiskDevice:
     description: Representation of one mounted device
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -13,13 +13,18 @@ $defs:
     type: object
     properties:
       major:
+        description: Major release of OS (aka the x version)
+        example: 6
         type: integer
         maxLength: 4
       minor:
+        description: Minor release of OS (aka the y version)
+        example: 8
         type: integer
         maxLength: 4
       name:
         description: Name of the distro/os
+        example: "RHEL"
         type: string
         enum: ["RHEL"]
   DiskDevice:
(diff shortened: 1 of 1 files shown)


iphands:
cc @ShimShtein

ShimShtein:
Sounds good, but I am a bit worry about enforcing enums. There always be someone with an exotic linux distro, and it looks like we won't want to reject the whole host because of this.

jharting:
@iphands more nesting means longer filter queries (e.g. `?filter[system_profile][operating_system][major][gte]=7`) which is not a a big deal, though

iphands:
> Sounds good, but I am a bit worry about enforcing enums. There always be someone with an exotic linux distro, and it looks like we won't want to reject the whole host because of this.

I think we do actually. We are specifically *not* aiming to open the Inventory to dump random host info into.
If there is not a business case for storing the data, we should drop.

beav (approved):

CodeHeeler (approved):
LGTM

schemas/system_profile/v1.yaml
```diff
@@ -9,6 +9,26 @@ $defs:
...
+      minor:
+        description: Minor release of OS (aka the y version)
+        example: 8
+        type: integer
+        maxLength: 4
+      name:
+        description: Name of the distro/os
+        example: RHEL
+        type: string
+        maxLength: 4
```
ryandillinfelton, line 30: minor nitpick, but maxLength feels redundant if we're enforcing an enum

iphands: WFM to remove I had seen another example somewhere else I think.

iphands: Or maybe even the linter complains. Lemme see

CodeHeeler: Yep, you're right, the linter does not like it if you do not include maxLength.

ryandillinfelton: Dang. We can just leave it in for now, it's not a big deal. @tahmidefaz I think we may want to adjust the linter a bit to account for stuff like this.

tahmidefaz: Yeah, it's doable :+1:

ryandillinfelton (approved):
This looks good to me

schemas/system_profile/v1.yaml
```diff
@@ -9,6 +9,26 @@ $defs:
...
+      major:
+        description: Major release of OS (aka the x version)
+        example: 6
+        type: integer
+        maxLength: 4
+      minor:
+        description: Minor release of OS (aka the y version)
+        example: 8
+        type: integer
+        maxLength: 4
```
tahmidefaz, line 25: Correct me if I am wrong, but, since `major` and `minor` are both `integer`s I think we should use `maximum` and `minimum`, yeah? I don't think `maxLength` doesn't do anything for type `integer`. :disappointed:

tahmidefaz (changes requested):

schemas/system_profile/v1.yaml
ryandillinfelton: Nope you're right. Good catch!

beav: @iphands Can you take a look at ^^ ? I think this is the last open comment

iphands: That wfm

beav:
lgtm as well

beav:
@tahmidefaz I think the PR is current at this point. can you take a look and re-review?

tahmidefaz:
@beav Sorry, our team was doing a hackathon the last sprint. Taking a look now! :+1:

tahmidefaz (approved):
LGTM :+1:

PR #29, opened 2020-11-24
Include all remaining packages not in installed_packages

CodeHeeler:
A way to have all packages passed along and not only the max of multi values provided by installed_packages without breaking those dependent on the current implementation of installed_packages.

Dependent on:
https://github.com/RedHatInsights/insights-puptoo/pull/69

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -263,6 +263,13 @@ $defs:
           example: "krb5-libs-0:-1.16.1-23.fc29.i686"
           type: string
           maxLength: 512
+      installed_packages_delta:
+        type: array  # sorted list, rest of packages not in installed_packages
+        items:
+          description: A NEVRA string for a single installed package
+          example: "krb5-libs-0:-1.16.1-23.fc29.i686"
+          type: string
+          maxLength: 512
       installed_services:
         type: array
         items:


kruai (commented):
This looks fine to me, though I wonder if we could define a regex "pattern" for both this field and installed_packages... Even something really loose like ^.+:.+$, just to make sure it's using the right format. That makes the assumption that all packages follow the format shown in the "example" field, though

CodeHeeler:
> This looks fine to me, though I wonder if we could define a regex "pattern" for both this field and installed_packages... Even something really loose like ^.+:.+$, just to make sure it's using the right format. That makes the assumption that all packages follow the format shown in the "example" field, though

Created an issue to track this works separately as we'll want to be careful with changes effecting installed_packages.
https://issues.redhat.com/browse/HBISPSC-27

Let's leave this PR as is. Thanks for the suggestion though!

kruai:
Thanks for creating a ticket to track that! I think this PR is fine, then.

kruai (approved):

CodeHeeler:
Updated name to installed_packages_delta as per today's discussion.

gmcculloug:
@iphands Can you take a quick look?  Should be ready to merge.

kruai (approved):

ryandillinfelton (approved):

thearifismail (approved):

ryandillinfelton (approved):

PR #30, opened 2020-12-01
add owner_id field [RHCLOUD-8583]

ryandillinfelton:
Mirror of changes needed for https://github.com/RedHatInsights/insights-host-inventory/pull/747/files.
Adds a new field to the sytem profile `owner_id` which is validated as a UUID


kruai (approved):

schemas/system_profile/v1.yaml
```diff
@@ -150,6 +150,9 @@ $defs:
     description: Representation of the system profile fields
     type: object
     properties:
+      owner_id:
+        type: string
+        pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
```
tahmidefaz, line 155: Can we have an example of how an `owner_id` might look like?

tahmidefaz (changes requested):

schemas/system_profile/v1.yaml
kruai: Good point, Tahmid... That would make a good reference, and help out the regex-impaired (like me)

ryandillinfelton: Added an example and a description

tahmidefaz:
@ryandillinfelton for some reason the pr check isn't working for this repo. I had some doubts, so I decided to run the lint check locally on my laptop, and I got:
`153:16  error  property-string-max-length  #/$defs/SystemProfile/properties/owner_id does not define max length`

thearifismail (approved):

ryandillinfelton:
Looks like the test is still busted. Closing and reopening [here](https://github.com/RedHatInsights/inventory-schemas/pull/34) to switch the check.

PR #31, opened 2020-12-01
operating_system needs inline def not ref

CodeHeeler:
To match the schema here, fixing a bug in CI due to use of ref in an array instead of bare, fix is to use inline def for now.

https://github.com/RedHatInsights/insights-host-inventory/pull/754


tahmidefaz:
@CodeHeeler apologies, for some reason our pr check isn't running in this repo. When I ran the lint I saw the following:
```/home/tefaz/Desktop/inventory-schemas/schemas/system_profile/v1.yaml
 168:24  error  property-description  #/$defs/SystemProfile/properties/operating_system does not define description
 168:24  error  property-example      #/$defs/SystemProfile/properties/operating_system does not define an example
 170:20  error  parser                Bad indentation of a mapping entry
 170:21  error  parser                Incomplete explicit mapping pair; a key node is missed
```
I don't think you need to worry about providing an `example`, it goes away when you fix the indentation.

tahmidefaz (approved):
LGTM :+1:

ryandillinfelton (approved):
LGTM

ryandillinfelton:
Will merge when Jenkins un-dies

iphands:
@kruai @gmcculloug can we go ahead and just merge this one?

tahmidefaz (approved):

kruai (approved):

PR #35, opened 2020-12-10
Fixing NestedObject definition

tahmidefaz:
I think the current `NestedObject` definition is that all the children of `options` must be objects. For example, a valid request is expected to look like the following: `options: {"uid": {"blah: {"another_one": {}}}}`. 

But, I think what we want `options` to look like is something like this: `options: {"uid":"0","ro":false, "random":{"yep":true}}`.
We can close/update this if this isn't accurate. :+1: 

Corresponding HBI PR: https://github.com/RedHatInsights/insights-host-inventory/pull/771

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -7,7 +7,10 @@ $defs:
     propertyNames:
       minLength: 1
     additionalProperties:
-      "$ref": "#/$defs/NestedObject"
+      oneOf:
+        - "$ref": "#/$defs/NestedObject"
+        - not:
+            type: object
   DiskDevice:
     description: Representation of one mounted device
     type: object
@@ -25,7 +28,6 @@ $defs:
         example:
           uid: "0"
           ro: true
-        type: object
         "$ref": "#/$defs/NestedObject"
       mount_point:
         description: The mount point
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -3,6 +3,7 @@ $id: system_profile.spec.yaml
 $schema: http://json-schema.org/draft-07/schema#
 $defs:
   NestedObject:
+    type: object
     description: An arbitrary object that doesn’t allow empty string keys.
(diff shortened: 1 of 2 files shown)


ryandillinfelton:
Is there no issue with having NestedObject not have a type? I thought only objects could have properties.

kruai:
That's a good point, @ryandillinfelton... I think we might have to use `anyOf` here, like...
```yml
NestedObject:
  type: object
  (...)
  additionalProperties:
    anyOf:
      - "$ref": "#/$defs/NestedObject"
      - type: string
```
Or something like that?

tahmidefaz:
Just updated it to use `oneOf` since this is an either/or scenario :+1:

kruai:
I like the way this is shaping up, but I'm still curious about @ryandillinfelton's point. Will NestedObject be able to have properties if it's not of type Object?

tahmidefaz:
In this current state, I don't think it's possible for `NestedObject` to have properties if it's not of type Object. Here, if the `additionalProperties` is not of type object, it'd add that to the object. If the additional property is an Object, it will recursively validate it with the `NestedObject` definition.

kruai (approved):

ryandillinfelton (approved):
Looks Gouda to me :cheese:

schemas/system_profile/v1.yaml
```diff
@@ -8,7 +8,10 @@ $defs:
     propertyNames:
       minLength: 1
     additionalProperties:
-      "$ref": "#/$defs/NestedObject"
+      oneOf:
+        - "$ref": "#/$defs/NestedObject"
+        - not:
+            type: object
```
Glutexo, line 14: This is smart, I like it! 👍🏻

Glutexo (commented):
Confirmed working! 👍🏻 I think this needs tests. I‘ll create a ticket for that.

Glutexo:
The ticket: https://issues.redhat.com/browse/RHCLOUD-11292

PR #36, opened 2020-12-11
added rhsm_releasever

MichaelMraka:
support for subscription-manager release lock

reopening this PR once again as it was incorrecly closed (see #18).

os release and rhsm release are independent values.

you can install e.g. RHEL 8.3 and set RHSM to lock version to 8.2

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -324,3 +324,12 @@ $defs:
         maxLength: 10
         description: The SELinux mode provided in the config file
         example: 'permissive'
+      rhsm:
+        description: Object for subscription-manager details
+        type: object
+        properties:
+          version:
+            description: System release set by subscription-manager
+            example: "8.1"
+            type: string
+            maxLength: 255


schemas/system_profile/v1.yaml
```diff
@@ -309,3 +309,8 @@ $defs:
         maxLength: 10
         description: The SELinux mode provided in the config file
         example: 'permissive'
+      rhsm_releasever:
+        description: System release version set by subscription-manager
+        type: string
+        example: "8.2"
+        maxLength: 100
```
kruai, line 316: 100 characters is really long for what I'd expect for this kind of field... Most cases I can think of aren't even longer than 10 chars. Was that intentional?

MichaelMraka: Well, it was copied from os_release field which should contain similar data. I agree it should not exceed 10 chars in foreseeable future.

iphands:
Hey can we split out the info like os.major os.minor here: https://github.com/RedHatInsights/inventory-schemas/blob/master/schemas/system_profile/v1.yaml#L174

Are there going to be other rhsm_ keys?
What if we make an rhsm dict/object and nest underneath that?

thearifismail:
Yes, rhsm dict/object is better but it should include `version`, like:
```yaml
rhsm-version:
  major or os_major:
  minor or os_minor:
```

beav:
If the client and server think two different things, who will win? For example:

- client is registered and checks in w/ both sub-man and insights-client
- on rhsm-web (or satellite perhaps), user sets RHSM rhel version to 8.2
- rh_cloud + yupana or rhsm-conduit runs
- user checks rhsm_releasever on system_profile
- sub-man runs on client
- insights-client upload happens again
- user checks rhsm_releasever on system_profile

It sounds like in this case, you'd get two different results for rhsm_releasever if it was always coming from the client. Is that expected?

iphands:
> Yes, rhsm dict/object is better but it should include `version`, like:
> 
> ```yaml
> rhsm-version:
>   major or os_major:
>   minor or os_minor:
> ```

Well... I would actually like to consider
```
rhsm:
  os_major:
  os_minor:
  ...
```

Grouping more things by RHSM... I think we anticipate more stuff here in the future, eh?
Or is that a bad assumption?

iphands:
> If the client and server think two different things, who will win? For example:
> 
> * client is registered and checks in w/ both sub-man and insights-client
> * on rhsm-web (or satellite perhaps), user sets RHSM rhel version to 8.2
> * rh_cloud + yupana or rhsm-conduit runs
> * user checks rhsm_releasever on system_profile
> * sub-man runs on client
> * insights-client upload happens again
> * user checks rhsm_releasever on system_profile
> 
> It sounds like in this case, you'd get two different results for rhsm_releasever if it was always coming from the client. Is that expected?

Wait Chris... I did not think that this was the reason for including the RHSM version.
If the RHSM version stuff is to settle a dispute about who is right I am going to try and close this.

If however a customer can have a different version form /etc/redhat-release and they can be validly independent then I'm okay with it.

I think this is what @MichaelMraka is suggesting.
What are the use cases that a customer would do this for here?
Is it something like: lock the RHSM tooling to an 8.2 version even though I update some RPMS to 8.3
Or is it something like: lock the release version to 8.2 and bring in a few RPMs from 8.3

What is actually going on here when things differ?

MichaelMraka:
> Grouping more things by RHSM... I think we anticipate more stuff here in the future, eh?
> Or is that a bad assumption?

I really don't know. What we (patch) need is to support "rhsm version lock" so we need only to know "releasever" set by rhsm.

>  What are the use cases that a customer would do this for here?
Is it something like: lock the RHSM tooling to an 8.2 version even though I update some RPMS to 8.3
Or is it something like: lock the release version to 8.2 and bring in a few RPMs from 8.3

Hard to say, I (as an admin) would register my system to z-stream if I do plan to consume security only updates.

> What is actually going on here when things differ?

Let's make it clear. We are talking about this:
``
[root@cd87be7d6699 /]# more /etc/redhat-release
Red Hat Enterprise Linux release 8.2 (Ootpa)
[root@cd87be7d6699 /]# subscription-manager release
Release not set
[root@cd87be7d6699 /]# dnf download zsh --url 
Updating Subscription Management repositories.
Last metadata expiration check: 0:00:26 ago on Wed Dec 16 09:36:59 2020.
https://cdn.redhat.com/content/dist/rhel8/8/x86_64/baseos/os/Packages/z/zsh-5.5.1-6.el8_1.2.x86_64.rpm
[root@cd87be7d6699 /]# subscription-manager release --list
+-------------------------------------------+
          Available Releases
+-------------------------------------------+
8
8.0
8.1
8.2
8.3
[root@cd87be7d6699 /]# subscription-manager release --set 8.1
Release set to: 8.1
[root@cd87be7d6699 /]#  dnf download zsh --url --disablerepo=local
Updating Subscription Management repositories.
Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)                  2.1 MB/s |  15 MB     00:07    
Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)                     2.1 MB/s |  15 MB     00:07    
Last metadata expiration check: 0:00:02 ago on Wed Dec 16 09:48:10 2020.
https://cdn.redhat.com/content/dist/rhel8/8.1/x86_64/baseos/os/Packages/z/zsh-5.5.1-6.el8_1.2.x86_64.rpm

``

Notice the `.../rhel8/8/...` vs. `.../rhel8/8.1/...` in the URLs. Which means these are two different repos with different set of packages. So when rhsm release is set we have to evaluate different updates for it.
OS release (aka /etc/redhat-release) is the same in both cases.

beav:
Ah i see, thanks for the additional info on the use case.

I would be ok with:

```
rhsm_client_config:
  rhsm_release: <whatever the version string is>
```
I'm pretty sure the main use of the string is to populate the `$version` in the dnf URL, so I don't think splitting it into major/minor is needed.

beav:
some example values from rhsm-web:

```
8
8.0
8.1
8.2
8.3
```

some example values from Satellite:
```
6Server                                                                                                                  
7Server
8
```

Both of the above also include the empty string.


If you make the change mentioned in https://github.com/RedHatInsights/inventory-schemas/pull/36#issuecomment-746293990 (maybe change rhsm_client_config to rhsm_client just for extra clarity), I think we can populate it via this patch and it should work:


```diff
--- src/puptoo/process.py       2021-01-04 18:37:01.002816106 -0500
+++ .venv/lib/python3.8/site-packages/src/puptoo/process.py     2021-01-04 18:37:52.100063722 -0500
@@ -8,6 +8,7 @@
 from insights.combiners.cloud_provider import CloudProvider
 from insights.combiners.redhat_release import RedHatRelease
 from insights.combiners.virt_what import VirtWhat
+from insights.combiners.rhsm_release import RhsmRelease
 from insights.combiners.sap import Sap
 from insights.core import dr
 from insights.parsers.cpuinfo import CpuInfo
@@ -53,6 +54,7 @@
         Specs.hostname,
         CpuInfo,
         VirtWhat,
+        RhsmRelease,
         MemInfo,
         IpAddr,
         DMIDecode,
@@ -82,6 +84,7 @@
     hostname,
     cpu_info,
     virt_what,
+    rhsm_release,
     meminfo,
     ip_addr,
     dmidecode,
@@ -193,6 +196,13 @@
             catch_error("virt_what", e)
             raise
 
+    if rhsm_release:
+        try:
+            profile["rhsm_client"] = {"rhsm_release": rhsm_release.set}
+        except Exception as e:
+            catch_error("rhsm_release", e)
+            raise
+
     if installed_rpms:
         try:
             profile["installed_packages"] = sorted(

```

I think it's OK to only read from the client. While crc could probably figure out the server-side value from rhsm-web, it would not be able to figure out the server-side value from the sat and we can't assume the satellite is also reporting stuff via the plugin. Also, you can create whatever release versions you want.

It looks like releasever is stored as a varchar(255) in candlepin so I would make this a 255 or 256 limit string, with empty allowed.

beav:
@MichaelMraka post-holiday ping; this PR should be ok to go w/ the changes above.

iphands:
@beav 

Hey I'd actually prefer:
```yaml
rhsm:
    release: String
```

Instead of rhsm_client_config and a nested rhsm_release.

iphands:
@MichaelMraka bump... we are talking about this in our weekly again.
Please engage so we can proceed, thanks!

MichaelMraka:
Hello @iphands and @beav, I'm sorry for the delay.
I do not have a strong opinion on naming. Just slightly prefer `rhsm: release: String`, it's shorter :) .

MichaelMraka:
If you agree I'll fix PR so we can move on.

beav:
> If you agree I'll fix PR so we can move on.

That works for me :) Are you able to say in the field description that it comes from the client? I just want it to be noted somewhere, so folks don't think it's the server-side version of the data.

MichaelMraka:
changed to `rhsm: release: String`

schemas/system_profile/v1.yaml
```diff
@@ -324,3 +324,12 @@ $defs:
...
         example: 'permissive'
+      rhsm:
+        description: Object for subscription-manager details
+        type: object
+        properties:
+          version:
+            description: System release set by subscription-manager
+            example: "8.1"
+            type: string
+            maxLength: 16
```
beav, line 335: This needs to be 255, since candlepin can have a 255 length str:

```java
    @Column(length = 255, nullable =  true)
    @Size(max = 255)
    private String releaseVer;
```

MichaelMraka: ok, extended to 255

beav (approved):

beav:
@kruai @tahmidefaz if you have time today, can you take a look at this PR? thanks!

kruai (approved):
This should be fine. I was going to request that we use a regex to only allow certain patterns in the `version` field, but based on the discussions above, that doesn't sound possible.

kruai:
Does this have an associated PR in https://github.com/RedHatInsights/insights-host-inventory ?

beav:
> Does this have an associated PR in https://github.com/RedHatInsights/insights-host-inventory ?

not yet to my knowledge. do we still need to make marshmallow updates to go along w/ schema updates?

kruai:
@beav Yeah - until we get HBI set up to pull the schema from inventory-schemas, the changes need to be made in both repos.

beav:
> @beav Yeah - until we get HBI set up to pull the schema from inventory-schemas, the changes need to be made in both repos.

https://github.com/RedHatInsights/insights-host-inventory/pull/808

image

tahmidefaz (changes requested):
Can you give us some examples of valid and invalid `rhsm` fields [in these two files](https://github.com/RedHatInsights/inventory-schemas/tree/master/tests/utils)? They are written in python, if you need any help, lmk! :slightly_smiling_face:

beav:
> Can you give us some examples of valid and invalid `rhsm` fields [in these two files](https://github.com/RedHatInsights/inventory-schemas/tree/master/tests/utils)? They are written in python, if you need any help, lmk! slightly_smiling_face

https://github.com/RedHatInsights/inventory-schemas/pull/46

I don't have access to this PR, but that one should cover it.

tahmidefaz (approved):
Thanks! :tada:

PR #37, opened 2020-12-15
Use x-propertyNames in place of propertyNames

tahmidefaz:
OpenApi specification does not support `propertyNames`. Check [here](https://swagger.io/docs/specification/data-models/keywords/).
We will use a custom validator to treat `x-propertyNames` as `propertyNames`.

HBI PR that handles `x-propertyNames`: https://github.com/RedHatInsights/insights-host-inventory/pull/771


gmcculloug:
@tahmidefaz If there is a PR for handling `x-propertyNames` it would be nice to include a link to that PR in the description so a reviewer can easily validate and help understand the dependency between the two PRs.

tahmidefaz:
@gmcculloug Good idea! Just added it! :+1:

kruai (approved):
LGTM, but we should wait until https://github.com/RedHatInsights/insights-host-inventory/pull/771 is merged first, right?

tahmidefaz:
Either way should be fine. But, we can wait for the HBI PR  :+1:

kruai:
Oh that's right, we still make changes in 2 places. In that case, we don't need to hold :)

ryandillinfelton (approved):
Looks good :+1:

PR #41, opened 2021-01-19
Add Pytest Action

tahmidefaz:
This PR is just to set up the pytest action. I will be removing the sample test and adding actual tests in a future PR.


kruai (approved):
I was gonna ask why we're using Ubuntu, but it looks like we were already using that for the linting job. Looks OK to me!

tahmidefaz:
> I was gonna ask why we're using Ubuntu, but it looks like we were already using that for the linting job. Looks OK to me!

Yes, that action is also using Ubuntu, and a good amount of the github actions example I looked at also used it as well.

ryandillinfelton (approved):
LGTM. I assume the empty requirements file will be get some content in the future?

PR #42, opened 2021-01-19
RHCLOUD-11292: Add Tests

tahmidefaz:
Jira: https://issues.redhat.com/browse/RHCLOUD-11292

The following tests were added:

- System Profile valid fields
- System Profile invalid fields
- Openapi Specification compatibility


.github/workflows/pr.yml
```diff
@@ -31,14 +31,17 @@ jobs:
...
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
         flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Validate openapi compatibility
+      run: |
+        openapi-spec-validator tests/helper-files/openapi.yaml
```
kruai, line 44: Are you running the validator against the helper file because it won't work against the raw System Profile schema?

tests/test_schema.py
```diff
@@ -0,0 +1,34 @@
+from openapi_spec_validator import validate_spec
+from jsonschema import Draft4Validator, Draft7Validator, RefResolver
+from jsonschema.validators import extend
+from jsonschema.exceptions import ValidationError
+import pytest
+from unittest import TestCase
+from yaml import safe_load
+
+from utils.invalids import INVALID_SYSTEM_PROFILES
+from utils.valids import VALID_SYSTEM_PROFILES
```
kruai, line 10: Ideally, we want to make this repo self-service for the stakeholders. Every time they make a change to the schema, these will need to be updated as well. Do you think they'll be comfortable making changes to those collections?

kruai (commented):
Everything looks good to me... I just had some comments I wanted to briefly discuss before giving my full approval.

.github/workflows/pr.yml
tahmidefaz: Yes, the standalone System Profile schema isn't a valid openapi schema. It must be paired with the minimal openapi spec file

tests/test_schema.py
tahmidefaz: I am not 100% sure if they will be comfortable making changes to those collections. If they are not, I think we can do it for them. In my opinion, if we test the schema here, we don't have to do so in HBI which allows us to catch potential issues/undesired behaviour before update the schema in HBI.

thearifismail (approved):
LGTM

PR #45, opened 2021-02-08
feat(sysprofile): Add satellite_id schema

SteveHNH:
Puptoo can send the satellite_id of a host and we currently do not have
this accounted for in the schema.

Signed-off-by: Stephen Adams <tsadams@gmail.com>


kruai:
This looks fine on a technical level, but I'm confused about why we're duplicating this info. `satellite_id` is already a field on the Host level that serves as a canonical fact. Why do we need it defined again in the system profile?

thearifismail (commented):
Needs a test or two for `satellite_id`.  Here is a sample that I plan to use for an Id, I need to add https://github.com/RedHatInsights/inventory-schemas/blob/c7124aefe3044bfb9a9292db8665cd0db9893c01/tests/utils/valids.py#L49

tahmidefaz:
Yes, @SteveHNH some tests in [here](https://github.com/RedHatInsights/inventory-schemas/tree/master/tests/utils), would be great!

kruai:
Please ignore the above comment - the bot's still WIP. It should definitely be running against more than 138 hosts :)

tahmidefaz (changes requested):
The schema changes look good to me. If you can add some `valid` and `invalid` cases in [these files](https://github.com/RedHatInsights/inventory-schemas/tree/master/tests/utils), that'd be great. Also, I am not sure if we are waiting on anyone here.

kruai:
> The schema changes look good to me. If you can add some `valid` and `invalid` cases in [these files](https://github.com/RedHatInsights/inventory-schemas/tree/master/tests/utils), that'd be great. Also, I am not sure if we are waiting on anyone here.

I think we're just waiting on those valid and invalid cases!

SteveHNH:
@CodeHeeler I came back to look at this and now I can't remember why we needed it. We have satellite ID in canonical facts. 

Do you happen to remember?

CodeHeeler:
I don't recall, but the other time in the team agenda it was mentioned was by csams on Feb 10.  It's not in any either of our Jira's.

lphiri:
@SteveHNH @gmcculloug @kruai  -  satellite_id has been  a source of confusion.  It is in fact always the same as the subscription manager id.     IMHO we should get rid of it altogether.

PR #49, opened 2021-02-16
rhc-11619: add "rhc_client_id" for supporting cloud_connector

thearifismail:
This PR adds a new field `rhc_client_id` to system profile for supporting cloud_connector.

The corresponding HBI PR is https://github.com/RedHatInsights/insights-host-inventory/pull/812


tests/utils/invalids.py
```diff
@@ -49,6 +49,8 @@
     {"selinux_current_mode": "random"},
     {"selinux_config_file": "random"},
     {"owner_id": "x"*12},
+    {"client_id": "x"*12},
```
tahmidefaz, line 52: Are you trying to check the maxLength here? Then this should be ` {"client_id": "x"*37}`.

tests/utils/invalids.py
```diff
@@ -49,6 +49,8 @@
...
+    {"client_id": "plxi13y1-99ut-3rdf-bc10-84opf904lfad"},
```
ryandillinfelton, line 53: I'd like some confirmation from @tahmidefaz that it's okay to have two invalid values in the test. It looks to me like it's fine though.

ryandillinfelton (approved):
LGTM. Just want some confirmation about the invalids from Tahmid when he reviews this.

tests/utils/invalids.py
tahmidefaz: Yeah, the more the better :+1:

tests/utils/invalids.py
ryandillinfelton: I interpreted it as trying to test that the pattern enforces a minimum length. Having both would probably be best though.

tahmidefaz: I assumed the second invalid test was checking the pattern enforcement

thearifismail: yes because the valid pattern take [0-9a-f]

CodeHeeler (approved):
LGTM, regex works as expected, thanks!

lphiri:
@thearifismail  - we agreed to use a more specific name here - I think we settled on "rhc_client_id"  to make it clear is specific to "RHC".

kruai:
Here are the System Profile validation results using the past 3 days of data.
Validating against the RedHatInsights/inventory-schemas master spec:
```
me:
	Pass:148
	Fail:0
	
```
Validating against this PR's spec:
```
me:
	Pass:148
	Fail:0
	
```

kruai:
I agree with Lindani; I also have "rhc_client_id" written down as the field name we agreed on.

kruai (approved):
LGTM :+1:

tahmidefaz (approved):

kruai:
FYI, here's the [HBI PR](https://github.com/RedHatInsights/insights-host-inventory/pull/812).

PR #51, opened 2021-03-02
Add smart_management_enabled field to system profile

kruai:
Addresses this Jira: https://issues.redhat.com/browse/RHCLOUD-11978
Adds the boolean "smart_management_enabled" field to the System Profile.
The corresponding host-inventory PR is here: https://github.com/RedHatInsights/insights-host-inventory/pull/822


ryandillinfelton (approved):

kruai:
It turns out this would have tracked the same data as the is_marketplace field, so we're not adding this.

PR #54, opened 2021-03-16
Change selinux_config_file from enum to allow a string

gmcculloug:
Fixes https://issues.redhat.com/browse/HBISPSC-38

Turns out sestatus code allows a value like `error (X)"` where X is more info about the error.
This change allows the `selinux_config_file` field to contain any string up to 128 characters.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -331,8 +331,7 @@ $defs:
         example: 'enforcing'
       selinux_config_file:
         type: string
-        enum: [enforcing, permissive, disabled]
-        maxLength: 10
+        maxLength: 128
         description: The SELinux mode provided in the config file
         example: 'permissive'
       is_marketplace:
--- a/tests/utils/invalids.py
+++ b/tests/utils/invalids.py
@@ -47,7 +47,7 @@ INVALID_SYSTEM_PROFILES = (
     {"sap_version": "1.00.122.04.147857563665464"},
     {"tuned_profile": "x"*257},
     {"selinux_current_mode": "random"},
-    {"selinux_config_file": "random"},
+    {"selinux_config_file": "x"*129},
     {"owner_id": "x"*12},
     {"rhc_client_id": "x"*12},
     {"rhc_client_id": "plxi13y1-99ut-3rdf-bc10-84opf904lfad"},


CodeHeeler (approved):
LGTM, thanks for making this change to account for the additional error option!

jharting:
@gmcculloug @beav is the error message going to be useful for any platform apps/services? In theory, an error may occur when collecting of any of the system profile facts. Yet I am not sure if these error belong to system profile. In general, IMO it's better to log the error and set the field to undefined than to keep errors/stacktraces/etc. in system profile.

beav:
> @gmcculloug @beav is the error message going to be useful for any platform apps/services? In theory, an error may occur when collecting of any of the system profile facts. Yet I am not sure if these error belong to system profile. In general, IMO it's better to log the error and set the field to undefined than to keep errors/stacktraces/etc. in system profile.

This particular issue is a bit strange. The `sestatus -b` command has a "mode from config file" field which maps to `selinux_config_file` on the system profile. There's also `selinux_current_mode` which reads from another field.

Here is the relevant snippet from the `sestatus` command for parsing:
```c
        printf_tab("Mode from config file:");
        if (selinux_getenforcemode(&rc) == 0) {
                switch (rc) {
                case 1:
                        printf("enforcing\n");
                        break;
                case 0:
                        printf("permissive\n");
                        break;
                case -1:
                        printf("disabled\n");
                        break;
                }
        } else {
                printf("error (%s)\n", strerror(errno));
        }
```

Normally I would expect `sestatus` to raise an error if it was unable to read the data, but it does not, which I'm assuming is an intentional choice they made.

Given that there are two modes set ("what is my actual current running status" and "what does sestatus think my conf file says"), I am OK with using a string here since it's what sestatus is reporting back.

beav:
@jharting does the response above cover your concerns?

beav:
we ended up removing the "error" string from being reporting via puptoo in https://github.com/RedHatInsights/insights-puptoo/pull/134.

If we want to relax the enforcement in the schema we can, or we can just rely on puptoo and close this PR.

kruai:
Apologies for all the dippy-bot comments. We're rolling out a fix now.

jharting:
@beav ah, I didn't notice we already have `selinux_current_mode` where the value semantics are preserved.

gmcculloug:
> If we want to relax the enforcement in the schema we can, or we can just rely on puptoo and close this PR.
@beav Thoughts on who is the best person to make that decision?  Seems like we are really close to wrapping this one up and it will help resolve a lot of validation errors.

beav:
> > If we want to relax the enforcement in the schema we can, or we can just rely on puptoo and close this PR.
> 
> @beav Thoughts on who is the best person to make that decision? Seems like we are really close to wrapping this one up and it will help resolve a lot of validation errors.

Let's kill this PR and go with the puptoo change

gmcculloug:
@beav @kruai @jharting Here is the part I am concerned about after discussing a little more with @beav.

We have two properties `selinux_current_mode` and `selinux_config_file`.  If they are both reported correctly they should be the same, if they are different that would likely mean the file was changed but the system has not rebooted.  (I think I am stating this correctly.)

The puptoo change (https://github.com/RedHatInsights/insights-puptoo/pull/134) is dropping "selinux_config_file" if the value is not one of the defined types which mean the system profile value will remain at the previous value.  Seems like this would be misleading.

Do others consider this a big enough concern?

cc @SteveHNH

beav:
> @beav @kruai @jharting Here is the part I am concerned about after discussing a little more with @beav.
> 
> We have two properties `selinux_current_mode` and `selinux_config_file`. If they are both reported correctly they should be the same, if they are different that would likely mean the file was changed but the system has not rebooted. (I think I am stating this correctly.)
> 
> The puptoo change ([RedHatInsights/insights-puptoo#134](https://github.com/RedHatInsights/insights-puptoo/pull/134)) is dropping "selinux_config_file" if the value is not one of the defined types which mean the system profile value will remain at the previous value. Seems like this would be misleading.
> 
> Do others consider this a big enough concern?
> 
> cc @SteveHNH

This came up after my prior reply. I think the use case of "I reported as 'enabled' in my conf file, then edited it with a typo but still see 'enabled'" is reason enough to do this PR instead of the other one.

I'm not aware of any unaddressed concerns at this point, this PR should be ok to merge.

gmcculloug:
@jharting Wanted to make sure you reviewed https://github.com/RedHatInsights/inventory-schemas/pull/54#issuecomment-802252649 and agreed with the suggestions being made there.  Otherwise, I think this PR is ready to merge.

Also note that https://github.com/RedHatInsights/insights-puptoo/pull/134 was merged, but I feel it should be reverted based on my previous comments.  cc @SteveHNH @CodeHeeler

jharting:
> @jharting Wanted to make sure you reviewed [#54 (comment)](https://github.com/RedHatInsights/inventory-schemas/pull/54#issuecomment-802252649) and agreed with the suggestions being made there. Otherwise, I think this PR is ready to merge.

Given the fact that we also have `selinux_current_mode` and, unlike that key, this one is supposed to reflect what's actually in the configuration file, your proposal sounds good to me.

gmcculloug:
After a quick discussion today this is ready to go if anyone wants the honor of approving/merging.  cc @kruai

CodeHeeler (approved):
Agreed, as per discussion.  LGTM.  Thanks!

kruai (approved):

kruai:
It looks like this needs one more review in order for us to merge it, but I give you permission to use your admin powers to merge @gmcculloug :stuck_out_tongue: 

Just remember to merge the HBI PR as well!

tahmidefaz (approved):
Got it!

PR #55, opened 2021-03-16
Update CODEOWNERS path and add codeheeler

kruai:
Does a couple of things:
1. Updates the System Profile spec folder path to reference the specific folder that we update (from root, not just any folder matching that pattern)
2. Adds @CodeHeeler to the list of reviewers, since she's a stakeholder

I'm not sure why Stephen Adams keeps getting added as a reviewer to every PR, but nobody else we've listed (besides HBI team) gets added. I'm hoping that fixing the path will get it to work

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,4 @@
 * @RedHatInsights/host-based-inventory-committers
 
 # Stakeholders and HBI devs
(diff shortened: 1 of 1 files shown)


tahmidefaz (approved):
Nice! Let's see what happens :crossed_fingers:

kruai:
Whoops, it still requires 2 reviews to merge, and I guess it didn't add the other team members (since you reviewed on behalf of the group)! CC @ryandillinfelton @thearifismail

gmcculloug (approved):

PR #56, opened 2021-03-22
[HBISPSC-52] Rename installed_packages_delta

CodeHeeler:
Customer feedback requested renaming from delta to multiple to
be more clear.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -283,7 +283,7 @@ $defs:
           example: "krb5-libs-0:-1.16.1-23.fc29.i686"
           type: string
           maxLength: 512
-      installed_packages_delta:
+      installed_packages_multiple:
         type: array  # sorted list, rest of packages not in installed_packages
         items:
           description: A NEVRA string for a single installed package
--- a/tests/utils/invalids.py
+++ b/tests/utils/invalids.py
@@ -69,5 +69,5 @@ INVALID_SYSTEM_PROFILES = (
     {"is_marketplace": "True"},
     {"yum_repos": [{"gpgcheck": "True"}]},
     {"yum_repos": [{"enabled": "False"}]},
-    {"installed_packages_delta": ["x" * 513]},
+    {"installed_packages_multiple": ["x" * 513]},
 )
--- a/tests/utils/valids.py
+++ b/tests/utils/valids.py
@@ -66,5 +66,5 @@ VALID_SYSTEM_PROFILES = (
     {"is_marketplace": True},
     {"yum_repos": [{"gpgcheck": True}]},
     {"yum_repos": [{"enabled": False}]},
-    {"installed_packages_delta": ["x" * 512]},
+    {"installed_packages_multiple": ["x" * 512]},
 )


CodeHeeler:
Related PRs:
https://github.com/RedHatInsights/insights-host-inventory/pull/841
https://github.com/RedHatInsights/insights-puptoo/pull/138

jharting:
Since system profile is part of [public API](https://cloud.redhat.com/docs/api/inventory#operations-hosts-api\.host\.get_host_system_profile_by_id) we should (ideally) avoid breaking changes such as renaming/removing fields.

kruai:
@CodeHeeler The naming is still confusing to me. If `installed_packages` is the list of currently-installed packages, what is `installed_packages_multiple` used for?

CodeHeeler:
I agree that the name is confusing no matter what and that changing the name causes complications; longterm, this entire field hopefully will go away once installed packages can contain everything without breaking other teams.

This was specifically requested by customers during the demo last week and this was the specific word choice they wanted that made sense for their usage.  I do not think any other teams are using this field except Drift and all we do is immediately recombine it with installed_packages and parse them all together, so the customers are the only drivers here.

CodeHeeler:
@kruai So installed_packages is not the full list of currently installed packages, it is a select subset where if there are multiple values installed for the same package, we make an educated guess which is the newest and most likely one they are using and send only that along and were dropping all the others previously.  Once Drift added the ability to handle multi value facts, we added installed_packages_delta as the delta between all the packages and the installed_packages subset.  installed_packages + installed_packages_delta = truly all the packages installed.  Hope that helps clarify.

jharting:
> @kruai So installed_packages is not the full list of currently installed packages, it is a select subset where if there are multiple values installed for the same package (for example, multiple kernels), we (in puptoo) make an educated guess which is the newest and most likely one they are using and send only that along and were dropping all the others previously. Once Drift added the ability to handle multi value facts, we added installed_packages_delta as the delta between all the packages and the installed_packages subset. installed_packages + installed_packages_delta = truly all the packages installed. Hope that helps clarify.

Might be good to mention some of that in the description of the field so that others do not have to wonder about this same thing.

> I do not think any other teams are using this field except Drift and all we do is immediately recombine it with installed_packages and parse them all together, so the customers are the only drivers here.

There is no way to know that. Any customer may build e.g. their automation scripts based on our public APIs and expect those to not change in a breaking way. Although it's unlikely they are depending on this particular field it's still a breaking change if they happen to. The field is already referenced e.g. [here](https://github.com/RedHatInsights/patchman-clients/blob/f2bc91a16d9310d4e149f06f326fe85207aad273/inventory/model_system_profile_spec_yaml_system_profile.go#L31) and [here](https://github.com/RedHatInsights/javascript-clients/blob/86c9750b870a6404af4b08b15f7282d610f0b35d/packages/host-inventory/api.ts#L1157) The [versioning guidelines](https://internal.cloud.redhat.com/docs/reference/rest/versioning/#semantic-versioning) deal with the topic of making breaking changes to public APIs

CodeHeeler:
@jharting Exactly, this is why versioning api's is important and I realize it's not ideal for there to be a name change at all.  We did our best on the original pr to come up with something that made sense to us, the delta between the full set and those listed in installed_packages, and yes we can add this explanation to the field description.  Customers have requested this change, but as you say, other customers may feel differently or have scripts that would be broken from the change, which is why we wanted to do this as soon as possible before more users adopt this.  I'm tagging in @jeromemarc to speak on priority for this from customers compared to effort/risk that you have pointed out.

CodeHeeler:
After further discussion on a call today, we decided that a better description and documentation clarifying that installed_packages is a duplicate free subset and those packages not included there, the delta, are in this field.  Long term, this field will be going away; it is not worth the added work to change the name for something that is not permanent.  Closing this and related PRs.  Thanks all for the in-depth discussion.

PR #58, opened 2021-04-01
Add gpg_pubkey array

csams:
GPG pubkeys are stored as an array of package names with RPM sorting applied.
The sort probably has no meaning for these synthetic packages, but it makes
order consistent.

Signed-off-by: Christopher Sams <csams@redhat.com>

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -277,17 +277,24 @@ $defs:
         type: string
         maxLength: 32
       installed_packages:
-        type: array  # technically a set, ordering is not important
+        type: array  # technically a set, ordered by RPM sorting algorithm
         items:
           description: A NEVRA string for a single installed package
           example: "krb5-libs-0:-1.16.1-23.fc29.i686"
           type: string
           maxLength: 512
       installed_packages_delta:
-        type: array  # sorted list, rest of packages not in installed_packages
+        type: array  # packages not in installed_packages, ordered by RPM sorting algorithm
         items:
           description: A NEVRA string for a single installed package
           example: "krb5-libs-0:-1.16.1-23.fc29.i686"
+          type: string
+          maxLength: 512
+      gpg_pubkeys:
+        type: array  # technically a set, ordered by RPM sorting algorithm
+        items:
+          description: A package name string of a single imported GPG pubkey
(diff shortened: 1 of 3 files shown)


CodeHeeler (approved):
LGTM

CodeHeeler:
Related PRs:
https://github.com/RedHatInsights/insights-host-inventory/pull/857
https://github.com/RedHatInsights/insights-puptoo/pull/145

schemas/system_profile/v1.yaml
```diff
@@ -277,19 +277,26 @@ $defs:
...
           example: "krb5-libs-0:-1.16.1-23.fc29.i686"
           type: string
           maxLength: 512
+      gpg_pubkeys:
+        type: array  # technically a set, ordered by RPM sorting algorithm
+        items:
+          description: A package name string of a single imported GPG pubkey
+          example: "gpg-pubkey-11111111-22222222"
+          type: string
+          maxLength: 512
```
ryandillinfelton, line 299: How specific is the GPG public key format? Would it be possible to make this more specific?

ryandillinfelton (changes requested):
I'm requesting changes because we need some `valid` and `invalid` cases added to the test files here: https://github.com/RedHatInsights/inventory-schemas/tree/master/tests/utils. 

I also had a question about if we can tighten up the validation, if not that's fine, I just wanted to check since I'm not all that familiar with the format. Thanks!

schemas/system_profile/v1.yaml
kruai: Also, we should add the field to `valids.py` and `invalids.py`

kruai: Also, we should add the field to `valids.py` and `invalids.py`

kruai (changes requested):

schemas/system_profile/v1.yaml
csams: The format is basically the same as for other packages. I can infer a bit of structure, but I don't think we gain much by encoding it.

I updated valids/invalids. Thanks for pointing those out!

kruai (approved):
LGTM!

tahmidefaz (approved):
Looks good! :+1:

ryandillinfelton (approved):

PR #60, opened 2021-04-26
Remove unused tools folder

kruai:
Removes the /tools folder, which is unreferenced and unused.


tahmidefaz (approved):
Looks good! I think we need to update our README under a separate PR, since we have dippy bot now. :+1:

thearifismail (approved):

PR #61, opened 2021-04-29
Add field to identify edge hosts

tahmidefaz:
Subsequent HBI PR: https://github.com/RedHatInsights/insights-host-inventory/pull/892

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -350,6 +350,9 @@ $defs:
       is_marketplace:
         description: Indicates whether the host is part of a marketplace install from AWS, Azure, etc.
         type: boolean
+      is_edge:
+        description: Indicates whether the host is an edge host.
+        type: boolean
       rhsm:
         description: Object for subscription-manager details
         type: object
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -350,9 +350,12 @@ $defs:
       is_marketplace:
         description: Indicates whether the host is part of a marketplace install from AWS, Azure, etc.
         type: boolean
-      is_edge:
-        description: Indicates whether the host is an edge host.
-        type: boolean
+      host_type:
+        type: string
+        enum: [edge]
+        maxLength: 4
+        description: Indicates the type of host.
+        example: 'edge'
       rhsm:
         description: Object for subscription-manager details
         type: object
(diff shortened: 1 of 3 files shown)


tests/utils/invalids.py
```diff
@@ -71,4 +71,5 @@
     {"yum_repos": [{"gpgcheck": "True"}]},
     {"yum_repos": [{"enabled": "False"}]},
     {"installed_packages_delta": ["x" * 513]},
+    {"is_edge": "True"},
```
thearifismail, line 74: valid and invalid both have the same value `True`, how?  Shouldn't `invalid` be some non-boolean value?

tahmidefaz: resolving this as we switched to enum. @thearifismail the one in `valid` was a true `boolean` and the one in `invalid` was a `string`

tahmidefaz:
Moving this conversation over from https://github.com/RedHatInsights/insights-host-inventory/pull/892

> @tahmidefaz "is_edge" seems too specific as primary field. Should we make this into more an enum of types (perhaps "host_type")? that would be me extensible i think.

@lphiri Switched to an enum as per our discussion. I only have `edge` as a possible `host_type` as I am not sure what other possible `host_type` options we will have.

If anyone have any more enum suggestions, lmk!

lphiri:
@tahmidefaz LGTM. We don’t have to  define the other typed right away.

lphiri (approved):

kruai (approved):
LGTM

ryandillinfelton (approved):

schemas/system_profile/v1.yaml
```diff
@@ -350,6 +350,12 @@ $defs:
       is_marketplace:
         description: Indicates whether the host is part of a marketplace install from AWS, Azure, etc.
         type: boolean
+      host_type:
+        type: string
+        enum: [edge]
+        maxLength: 4
+        description: Indicates the type of host.
```
gmcculloug, line 357: This PR was merged while I was typing but still want to ask this question. 🙂   

Should the `description` indicate that this value is derived from other system properties, and not a value explicitly set on the host?  I think most properties here are explicit on the host which makes this one a little different.  @csams thoughts?

csams:
@tahmidefaz @kruai We need one more host_type in the enum to signify a type for existing systems.

PR #69, opened 2021-06-09
Mac address validation

ryandillinfelton:
This PR is to address this [Jira](https://issues.redhat.com/browse/RHCLOUD-14534)

Essentially we need some stricter validation for the MAC address. You might notice that the regex is pretty extensive. It is that way to validate all of the possible valid MAC address formats, including those weird 20 byte ones that infiniband uses. 

I've prepared a [regex101 example](https://regex101.com/r/BO5OEz/1) so you can confirm that it works as expected


tests/utils/valids.py
```diff
@@ -1,7 +1,16 @@
...
+    {"network_interfaces": [{"mac_address": "BD0DC5FB4235"}]},
+    {"network_interfaces": [{"mac_address": "d3a94b06bbdd"}]},
+    {"network_interfaces": [{"mac_address": "99:40:16:A9:38:21"}]},
+    {"network_interfaces": [{"mac_address": "2f:3c:00:53:8c:71"}]},
+    {"network_interfaces": [{"mac_address": "58-CA-D4-5F-D6-BE"}]},
+    {"network_interfaces": [{"mac_address": "d5-90-c8-e0-3b-e5"}]},
+    {"network_interfaces": [{"mac_address": "1EDC.C1E7.32BA"}]},
+    {"network_interfaces": [{"mac_address": "a2da.8b79.40e0"}]},
+    {"network_interfaces": [{"mac_address": "00:11:22:33:44:55:66:77:88:99:aa:bb:cc:dd:ee:ff:00:11:22:33"}]},
+    {"network_interfaces": [{"mac_address": "00112233445566778899aabbccddeeff00112233"}]},
```
kruai, line 13: These are all valid? I've only seen the ones with colons separating sets of 2 characters :O

tahmidefaz:
Looking at the slight increase in failures reported by `dippy-bot`... I feel like puptoo maybe sending us invalid `mac_addresses` :thinking:

tests/utils/valids.py
ryandillinfelton: Yep, there are a few allowed delimiters. The one with groups of four is a Cisco standard, but it's common enough I found it so I figured we should support it.

ryandillinfelton:
> Looking at the slight increase in failures reported by `dippy-bot`... I feel like puptoo maybe sending us invalid `mac_addresses`

I'm curious what they look like. Maybe they're sending something odd like an empty string or something when they have no value for it.

tests/utils/valids.py
gmcculloug: @roliveri Can you give this a quick review?

kruai:
> > Looking at the slight increase in failures reported by `dippy-bot`... I feel like puptoo maybe sending us invalid `mac_addresses`
> 
> I'm curious what they look like. Maybe they're sending something odd like an empty string or something when they have no value for it.

The validator leaves logs for these failures! [This kibana query](https://kibana.apps.crcp01ue1.o9m8.p1.openshiftapps.com/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2021-06-10T00:00:00.000Z',to:'2021-06-10T02:00:00.000Z'))&_a=(columns:!(_source),filters:!(),index:'43c5fed0-d5ce-11ea-b58c-a7c95afd7a5d',interval:auto,query:(language:kuery,query:'@message:%20%22Message%20failed%20validation%22'),sort:!())) should do it, although you'll need to drill down a bit to see those specific failures

roliveri (approved):
Looks good to me.
Has it been verified that all the additional failures are due to invalid data that is being detected by the new spec?

ryandillinfelton:
@kruai @tahmidefaz @SteveHNH How do we feel about merging this as is? We know it'll cause some failures, but they are sending invalid values. How likely would it be to get a fix for puptoo in this sprint?

kruai (approved):
Yep @ryandillinfelton I agree with merging this. I do think Stephen should weigh in first, though.
@SteveHNH - thoughts?

tests/utils/valids.py
kruai: Resolving this since Rich left his approving review.

SteveHNH:
@kruai @ryandillinfelton I think I need to do some mac address validation in puptoo to address this. At least to catch those 0.0.0.0 ones. Not even sure why a machine would send those. I'll throw something together now and get it fixed up

SteveHNH:
https://github.com/RedHatInsights/insights-puptoo/pull/171

ryandillinfelton:
@tahmidefaz could I get a review from you? I need one more approval to merge. Stephen and I are working together on a release strategy so that side of things is all worked out now.

tahmidefaz (approved):
Here you go! :+1:

PR #71, opened 2021-06-13
Add deployments from rpm ostree status to system profile.

csams:
Signed-off-by: Christopher Sams <csams@redhat.com>


schemas/system_profile/v1.yaml
```diff
@@ -131,6 +131,48 @@ $defs:
...
+    properties:
+      id:
+        description: ID of the deployment
+        example: "fedora-silverblue-63335a77f9853618ba1a5f139c5805e82176a2a040ef5e34d7402e12263af5bb.0"
+        type: string
+        minLength: 1
+        maxLength: 256
+      checksum:
+        description: The checksum / commit of the deployment
+        example: "63335a77f9853618ba1a5f139c5805e82176a2a040ef5e34d7402e12263af5bb"
```
ryandillinfelton, line 147: Is this hex? We could use regex here to insure that it's valid hex AND specifically 64 characters long.

tests/test_schema.py
```diff
@@ -1,4 +1,4 @@
-from openapi_spec_validator import validate_spec
+# from openapi_spec_validator import validate_spec
```
ryandillinfelton, line 1: if you don't need this could you delete the whole line please.

tests/utils/invalids.py
```diff
@@ -81,4 +81,39 @@
     {"greenboot_status": "blue"},
     {"greenboot_status": "too long"},
     {"greenboot_fallback_detected": "x"},
+    {"rpm_ostree_deployments": [{  # checksum empty
```
ryandillinfelton, line 84: This testing cold be more thorough. You only verify the checksum validation here.

ryandillinfelton (changes requested):
Requesting:

- [ ] Stricter checksum validation
- [ ] more invalid examples
- [ ] some cleanup

gmcculloug:
@beav Can you please review the OSTree properties and structure here and verify it makes sense?

@csams Do you think we'll need to be able to search on any of these properties?

tests/test_schema.py
csams: Sure, I'll delete it. Tests won't even run with the line, so I'm not sure how anyone is testing this repo with it as-is.

schemas/system_profile/v1.yaml
csams: Sounds good.

tests/utils/invalids.py
csams: done

csams:
@gmcculloug We'll need to make the checksum searchable.

beav:
> @gmcculloug We'll need to make the checksum searchable.

Do you just need "matches first part of string" or full substring?

Like if I had: `1234abc` and `34a777`, if I searched for `34a` would you expect one result or 2 results?

csams:
@beav @gmcculloug We can defer the search capability for now if it can be added later.

ryandillinfelton (approved):
looks good now. Thanks Chris!

kruai (approved):
LGTM

PR #77, opened 2021-09-27
Combine system_purpose fields into an object

kruai:
This is another PR addressing [ESSNTL-600](https://issues.redhat.com/browse/ESSNTL-600). Ronny noted that these fields would make more sense on a combined object, so I went ahead and made that change.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -440,21 +440,25 @@ $defs:
             example: "8.1"
             type: string
             maxLength: 255
-      system_purpose_usage:
-        description: The intended usage of the system
-        type: string
-        maxLength: 17
-        enum: ['Production', 'Development/Test', 'Disaster Recovery']
-        example: 'Production'
-      system_purpose_role:
-        description: The intended role of the system
-        type: string
-        maxLength: 37
-        enum: ['Red Hat Enterprise Linux Server', 'Red Hat Enterprise Linux Workstation', 'Red Hat Enterprise Linux Compute Node']
-        example: 'Red Hat Enterprise Linux Server'
-      system_purpose_sla:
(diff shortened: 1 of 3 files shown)


schemas/system_profile/v1.yaml
```diff
@@ -440,21 +440,25 @@ $defs:
...
-        example: 'Premium'
+      system_purpose:
+        description: Object for system purpose information
+        type: object
+        properties:
+          usage:
+            description: The intended usage of the system
+            type: string
+            maxLength: 17
+            enum: ['Production', 'Development/Test', 'Disaster Recovery']
```
thearifismail, line 451: How about changing 'Development/test' to `devtest`, which used to be a common term for this purpose.

kruai: The values aren't up to us. More info here: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_an_advanced_rhel_installation/configuring-system-purpose-advanced_installing-rhel-as-an-experienced-user

RonnyPfannschmidt: @kruai so the question would be on whether it makes sense as subfield for the rhsm data?

thearifismail (approved):
LGTM

schemas/system_profile/v1.yaml
kruai: Are you saying these fields are only applicable to RHSM? If so, then I get where you're coming from... but otherwise I'd say we should leave it separate

RonnyPfannschmidt: @kruai indeed, right now i dont know, otherwise we might still rename from system_purpose to rhel_system_purpose

ryandillinfelton (approved):

PR #78, opened 2021-09-29
Generic filtering xjoin related changes

ryandillinfelton:
Generic filtering for xjoin work has lead to some small changes in how the schema should be. I've updated the schema to be compliant and updated our instructions for contribution to include the new requirements.

The biggest change is the need to provide an example string with 3 different (unless that's impossible) example values for new string fields. This is so that we can automatically populate example hosts in xjoin-search with realistic data for manual and automated tests.

The requirements may still change as test generation work for generic filtering on xjoin is ongoing. Please let me know if there are any concerns around the changes or there are any better ways to insure compliance with the new requirements, especially the example strings.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -103,14 +103,14 @@ $defs:
         type: array
         items:
           description: The ipv4 address of the system
-          example: "123.456.789.012 123.456.789.012 123.456.789.012, 123.456.789.012 123.456.789.012, 123.456.789.012"
+          example: "227.161.169.210 30.143.76.153 233.87.178.7, 60.209.47.155 40.124.217.134, 67.77.119.70"
           type: string
           format: ipv4
       ipv6_addresses:
         type: array
         items:
           description: The ipv6 address of the system
-          example: "0123:4567:89ab:cdef:0123:4567:89ab:cdef , abcd:4567:89ab:cdef:0123:4567:89ab:cdef , qwer:4567:89ab:cdef:0123:4567:89ab:cdef"
+          example: "8886:2565:f753:1bbb:1d08:4239:c470:a889, dd2e:879f:afff:7845:b346:bb88:bcf2:4b1b, e979:3081:7218:4c98:fd19:5777:309a:957b"
           type: string
           format: ipv6
       mtu:
@@ -178,9 +178,11 @@ $defs:
       booted:
         description: Whether the deployment is currently booted
         type: boolean
+        example: true
       pinned:
         description: Whether the deployment is currently pinned
         type: boolean
+        example: false
   SystemProfile:
     title: SystemProfile
     description: Representation of the system profile fields
@@ -447,13 +449,13 @@ $defs:
         enum: [edge]
         maxLength: 4
         description: Indicates the type of host.
-        example: 'edge, not_edge, edge'
+        example: 'edge, edge, edge'
       greenboot_status:
         type: string
         enum: [red, green]
         maxLength: 5
         description: Indicates the greenboot status of an edge device.
-        example: 'green, red, green'
+        example: 'green, red, purple'
       greenboot_fallback_detected:
         type: boolean
         description: Indicates whether greenboot detected a rolled back update on an edge device.
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -19,12 +19,13 @@ $defs:
     type: object
     properties:
       device:
-        example: "/dev/fdd0"
+        example: "/dev/fdd0, /dev/fdd1, /dev/fdd2"
         type: string
         maxLength: 2048
       label:
         description: User-defined mount label
         type: string
+        example: "foo, bar, baz"
         maxLength: 1024
       options:
         description: Mount options for nested object
@@ -32,14 +33,15 @@ $defs:
           uid: "0"
           ro: true
         "$ref": "#/$defs/NestedObject"
+        x-indexed: false
       mount_point:
         description: The mount point
-        example: "/mnt/remote_nfs_shares"
+        example: "/mnt/remote_nfs_shares, /mnt/local_nfs, /mnt/foo"
         type: string
         maxLength: 2048
       type:
         description: The mount type
-        example: "ext3"
+        example: "ext1, ext2, ext3"
         type: string
         maxLength: 256
   YumRepo:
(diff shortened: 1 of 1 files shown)


schemas/system_profile/v1.yaml
```diff
@@ -139,41 +147,39 @@ $defs:
...
       version:
         description: The version of the deployment
-        example: "33.21"
+        example: "33.21, 31.12, 33.45"
         type: string
         minLength: 1
         maxLength: 255
       booted:
         description: Whether the deployment is currently booted
-        example: True
```
kruai, line 172: examples are required for each field, even if they're boolean, right?

schemas/system_profile/v1.yaml
```diff
@@ -420,13 +443,13 @@ $defs:
         enum: [edge]
         maxLength: 4
         description: Indicates the type of host.
-        example: 'edge'
+        example: 'edge, not_edge, edge'
```
kruai, line 446: `edge` is the only allowed value right now

schemas/system_profile/v1.yaml
```diff
@@ -420,13 +443,13 @@ $defs:
...
       greenboot_status:
         type: string
         enum: [red, green]
         maxLength: 5
         description: Indicates the greenboot status of an edge device.
-        example: 'red'
+        example: 'green, red, green'
```
kruai, line 452: `green` is listed twice

kruai (changes requested):

schemas/system_profile/v1.yaml
```diff
@@ -112,25 +120,25 @@ $defs:
         maximum: 18446744073709551615
       mac_address:
         description: MAC address (with or without colons)
-        example: "00:00:00:00:00:00"
+        example: "00:00:00:00:00:00, 10:00:00:00:00:00, 20:00:00:00:00:00"
         type: string
-        pattern: "^([A-Fa-f0-9]{2}[:-]){5}[A-Fa-f0-9]{2}$|^([A-Fa-f0-9]{4}[.]){2}[A-Fa-f0-9]{4}$|^[A-Fa-f0-9]{12}$|^([A-Fa-f0-9]{2}[:-]){19}[A-Fa-f0-9]{2}$|^[A-Fa-f0-9]{40}$"
+        pattern: "^([A-Fa-f0-9]{2}[:-]){5}[A-Fa-f0-9]{2}$,^([A-Fa-f0-9]{4}[.]){2}[A-Fa-f0-9]{4}$,^[A-Fa-f0-9]{12}$,^([A-Fa-f0-9]{2}[:-]){19}[A-Fa-f0-9]{2}$,^[A-Fa-f0-9]{40}$"
```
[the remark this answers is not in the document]
ryandillinfelton: Nope, switched it back

schemas/system_profile/v1.yaml
ryandillinfelton: I need some kind of alternative for the sake of testing filtering functionality so for enums with only one option I have to use a fake option. Validity isn't a huge concern since that validation is done on ingestion and doesn't interfere with the filtering test. I suppose it could be misleading to those reading the schema though so I'll look into alternatives.

schemas/system_profile/v1.yaml
ryandillinfelton: I understand it's redundant but it's great for the tests.

ryandillinfelton: Changed :+1:

kruai (commented):
Almost ready! Just one more Q

schemas/system_profile/v1.yaml
```diff
@@ -420,13 +449,13 @@ $defs:
         enum: [edge]
         maxLength: 4
         description: Indicates the type of host.
-        example: 'edge'
+        example: 'edge, edge, edge'
```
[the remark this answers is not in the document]
kruai: OK. It looks strange, and might be a little weird for end-users, but it's not incorrect. I can go ahead and approve.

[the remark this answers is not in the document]
ryandillinfelton: I'll look into changing the method's down the line but it will require some modification to the updater script.

schemas/system_profile/v1.yaml
```diff
@@ -112,25 +120,25 @@ $defs:
         maximum: 18446744073709551615
       mac_address:
         description: MAC address (with or without colons)
-        example: "00:00:00:00:00:00"
+        example: "00:00:00:00:00:00, 10:00:00:00:00:00, 20:00:00:00:00:00"
```
thearifismail, line 123: at least one example should be without colons

schemas/system_profile/v1.yaml
```diff
@@ -19,27 +19,29 @@ $defs:
     type: object
     properties:
       device:
-        example: "/dev/fdd0"
+        example: "/dev/fdd0, /dev/fdd1, /dev/fdd2"
```
thearifismail, line 22: need some variety in disk devices.

schemas/system_profile/v1.yaml
```diff
@@ -19,27 +19,29 @@ $defs:
...
         type: string
         maxLength: 2048
       label:
         description: User-defined mount label
         type: string
+        example: "foo, bar, baz"
```
thearifismail, line 28: Should be some realistic names.  foo, bar, baz sound like function names.

schemas/system_profile/v1.yaml
```diff
@@ -271,27 +286,32 @@ $defs:
...
         maxLength: 20
         pattern: '^\d+\.\d+\.\d+(\.\d+)?$'
         description: The kernel version represented with a three, optionally four, number scheme.
-        example: '3.10.0'
+        example: '4.18.2, 4.5.0, 5.1.0'
+        x-wildcard: true
       arch:
         type: string
         maxLength: 50
+        example: "ARM, x86_64, null"
```
thearifismail, line 308: Can an arch value be "null"?

schemas/system_profile/v1.yaml
ryandillinfelton: What would be a realistic name?

schemas/system_profile/v1.yaml
```diff
@@ -303,19 +323,23 @@ $defs:
...
         maxLength: 100
+        example: "ex1, ex2, ex3"
       katello_agent_running:
         type: boolean
       satellite_managed:
         type: boolean
       cloud_provider:
         type: string
         maxLength: 100
+        example: "aws, ms, ibm"
```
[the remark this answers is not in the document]
ryandillinfelton: We don't need every possibility to be an example

kruai (approved):

schemas/system_profile/v1.yaml
thearifismail: Something from "ls -l /dev" would be better but not critical

thearifismail (approved):

PR #79, opened 2021-10-06
Add ansible booleans to system profile

kruai:
Adds booleans to represent `ansible_controller_system`, `ansible_hub_system`, `ansible_catalog_worker_system`, and `ansible_sso_system`. Similar to how we handled `system_purpose`, I've combined them into an `ansible` object which has the boolean fields:

- controller_system
- hub_system
- catalog_worker_system
- sso_system

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -462,3 +462,23 @@ $defs:
             maxLength: 12
             enum: ['Premium', 'Standard', 'Self-Support']
             example: 'Premium'
+      ansible:
+        description: Object containing data specific to Ansible Automation Platform
+        type: object
+        properties:
+          controller_system:
+            description: The ansible-tower or automation-controller RPM is present on the host
+            type: boolean
+            example: True
+          hub_system:
+            description: The automation-hub RPM is present on the host
+            type: boolean
+            example: False
+          catalog_worker_system:
+            description: The catalog-worker RPM is present on the host
+            type: boolean
+            example: True
+          sso_system:
+            description: Host is an ansible-sso system
+            type: boolean
+            example: False
--- a/tests/utils/invalids.py
+++ b/tests/utils/invalids.py
@@ -243,4 +243,10 @@ INVALID_SYSTEM_PROFILES = (
         "role": "bar",
         "sla": "baz",
     }},
+    {"ansible": {
+        "controller_system": "strings are invalid",
+        "hub_system": 17,
+        "catalog_worker_system": "still invalid",
+        "sso_system": 1
+    }},
 )
--- a/tests/utils/valids.py
+++ b/tests/utils/valids.py
@@ -114,4 +114,10 @@ VALID_SYSTEM_PROFILES = (
         "role": "Red Hat Enterprise Linux Compute Node",
         "sla": "Self-Support"
     }},
+    {"ansible": {
+        "controller_system": True,
+        "hub_system": False,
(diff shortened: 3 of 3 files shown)


beav (approved):

jeromemarc:
@kruai The main reason is to be consistent with other workloads (e.g. SAP).

kruai:
> @kruai The main reason is to be consistent with other workloads (e.g. SAP).

@jeromemarc but is there any benefit of storing both? 
For example... is it possible to have a host with `ansible_controller_system = False` but `ansible_controller_version=3.8.3`?
Or, a host with `ansible_controller_system = True` but no `ansible_controller_version` set?

I ask because we really need to keep the size of System Profile fields down, and if we can add 50% the fields to track 100% of the information, I'd prefer to go that way

beav:
> I updated this to use the version fields instead of the booleans. Let me know what you think; I can revert to booleans if y'all decide we absolutely need both the existence booleans and version numbers



> > @kruai The main reason is to be consistent with other workloads (e.g. SAP).
> 
> @jeromemarc but is there any benefit of storing both? For example... is it possible to have a host with `ansible_controller_system = False` but `ansible_controller_version=3.8.3`? Or, a host with `ansible_controller_system = True` but no `ansible_controller_version` set?
> 
> I ask because we really need to keep the size of System Profile fields down, and if we can add 50% the fields to track 100% of the information, I'd prefer to go that way

Normally I am in favor of being verbose. Is there a max recommended size for system profile? I don't mind if we need to start shrinking it but it would be good to have a size target to work towards.

kruai:
OK @beav @jeromemarc @gmcculloug this PR now has all 8 fields and is ready for re-review

ryandillinfelton:
> > > I updated this to use the version fields instead of the booleans. Let me know what you think; I can revert to booleans if y'all decide we absolutely need both the existence booleans and version numbers
> > 
> > 
> > > > @kruai The main reason is to be consistent with other workloads (e.g. SAP).
> > > 
> > > 
> > > @jeromemarc but is there any benefit of storing both? For example... is it possible to have a host with `ansible_controller_system = False` but `ansible_controller_version=3.8.3`? Or, a host with `ansible_controller_system = True` but no `ansible_controller_version` set?
> > > I ask because we really need to keep the size of System Profile fields down, and if we can add 50% the fields to track 100% of the information, I'd prefer to go that way
> > 
> > 
> > Normally I am in favor of being verbose. Is there a max recommended size for system profile? I don't mind if we need to start shrinking it but it would be good to have a size target to work towards.
> 
> No, we don't have a size target. We already hit an initial limit of sorts where the maximum Kafka message size had to be increased, which comes with some performance implications... but I don't want to hold this back any longer, since people are waiting on it. I'll push a change with all 8 new fields

I'm all for being verbose, but I don't think the trade off is worth it in this case.

- We've had issues pop up in the past with reporters only setting part of fields that are "paired" like these are, so the risk of discrepancy is a bigger worry to me. If we only query on the ID there is no way for two queries to return conflicting results.
- We also have to consider the amount of data we are indexing in elastic-search for use in filters. If we want to use both of these fields for filtering we must index them both, it seems like a waste to index twice as many fields for the same functionality, even if elastic-search is quite performant with booleans. I can't say for sure what the impact of adding them is, but since the generic filtering work is already expanding the number of fields indexed drastically I'm inclined to add fewer where possible

I also don't think that a not null query is a particularly uncommon or ambiguous, however I'll cede that that's a matter of opinion.

PR #80, opened 2021-11-02
Add wildcard support for Ansible workload fields

kruai:
Sets up wildcard filter support for Ansible workload fields.


kruai:
This only affects how HBI processes the fields, so I'm only going to wait on HBI review :)

thearifismail (approved):

PR #83, opened 2021-11-18
Fix ipv4 address example

kruai:
"123.456.789.012" is not a valid ipv4 address. I've updated our example values to use "192.0.2.146", which _is_ valid.


ryandillinfelton (approved):

thearifismail (approved):
LTGM

PR #84, opened 2021-11-19
Require all operating_system fields

kruai:
@ryandillinfelton brought up a good point: if we're provided `operating_system` data, all 3 fields should be provided (name, major, and minor). It doesn't make sense to provide one field but not the others.


schemas/system_profile/v1.yaml
```diff
@@ -252,6 +252,7 @@ $defs:
       operating_system:
         description: Object for OS details. Supports range operations
         type: object
+        required: ["name"]
```
ryandillinfelton, line 255: I really think we should require all three. The main point in creating a structured operating_system object to have in addition to the unstructured `os_release` was so that we could know the format for sure. Having optional field seems like it undercuts that goal.

kruai: TBH I'd prefer that as well, as it would lead to fewer potential issues down the line. I think the reason these were originally all left optional was to allow for flexibility with non-RHEL systems, in case it could only supply an OS name, or name and major version, or something like that.

kruai: But bc I'd prefer they all be required, I'll go ahead and push that change to this PR. I can revert that one change if someone disagrees

CodeHeeler (approved):
LGTM

PR #85, opened 2021-11-23
Add CentOS to OS name enum

kruai:
Addresses [ESSNTL-1756](https://issues.redhat.com/browse/ESSNTL-1756). Adds "CentOS" to the `operating_system.name` enum, which will allow these systems to be processed by HBI.


CodeHeeler (approved):
LGTM.  I'm still not a huge fan of the enum in general, but since we're just sticking with what we already have, this works great, thank you!

PR #86, opened 2022-02-07
SPM-1324: add client updates data into SystemProfile

MichaelMraka:
package_updates is a subset of VMaaS UpdatesV2Response schema defined in
https://github.com/RedHatInsights/vmaas/blob/master/vmaas/webapp/webapp.v3.spec.yaml


kruai:
Based on our discussion on [SPM-1324](https://issues.redhat.com/browse/SPM-1324):
Because there is no need to store this data on the host, the plan is to send it via `platform_metadata.custom_metadata` instead.

PR #92, opened 2022-08-22
[ESSNTL-2327] Limit int values' max size

CodeHeeler:
Large values yield a 500 error but should be beyond a logical size for the fields so adding a max value to each of the int and int64 fields should not impact customers yet should protect against the potential for these errors.

Added the maximum value to the schema and updated the invalid values from the test utils.

Accompanying PRs to insights-host-inventory and xjoin-search will be added.

We should run a check to see how many hosts would be potentially effected by this change and if more drastic than expected, we could alternatively adjust filtering.


CodeHeeler:
Related xjoin-search PR:
https://github.com/RedHatInsights/xjoin-search/pull/163

FabriciaDinizRH (approved):

schemas/system_profile/v1.yaml
```diff
@@ -117,7 +117,7 @@ $defs:
         description: MTU (Maximum transmission unit)
         type: integer
         minimum: 0
-        maximum: 18446744073709551615
+        maximum: 2147483647
```
kruai, line 120: For all of the regular `integer` type fields, we should be able to just add `format: int32`. This sets the field's maximum to the signed 32-bit int's max value. More info in the "Numbers" section of [these docs](https://swagger.io/docs/specification/data-models/data-types/).

schemas/system_profile/v1.yaml
```diff
@@ -214,20 +214,20 @@ $defs:
...
         type: integer
         minimum: 0
-        maximum: 4294967295
+        maximum: 2147483647
       system_memory_bytes:
         type: integer
         format: int64
         minimum: 0
-        maximum: 18446744073709551615
+        maximum: 9007199254740991
```
kruai, line 230: For this one, I couldn't find a cleaner solution. `int64` format allows values up to 2^63 - 1, which is above the MAX_SAFE_INTEGER limit in JavaScript. It'd be cool if we could define a constant for this value, if possible, though I don't know if it is. If not, this is fine as-is.

kruai (commented):
Looks good, but I just wanted to suggest one change. The other comment is an optional request

schemas/system_profile/v1.yaml
CodeHeeler: Do you want this "in addition to" the min/max already present, like we have on our int64 format field, or were you saying "instead of" here?

CodeHeeler: i.e. even before this PR, we had min/max added everywhere; are you suggesting we remove those entirely and rely only on the stated format.

schemas/system_profile/v1.yaml
CodeHeeler: Attempted using the constant something like [this](https://stackoverflow.com/questions/68094841/is-it-possible-to-define-constants-in-yml-files/68094990#68094990), linter seemed to not find a colon implying it didn't work properly.  After discussing in slack, decided to just go with the suggestion to add a comment only for now.  Let me know if this is sufficient.

schemas/system_profile/v1.yaml
kruai: oh, I meant "instead of"; I'm pretty sure if we set the format to int32, it sets the max to the correct value.

schemas/system_profile/v1.yaml
kruai: I think the current comment is clear and helpful.

schemas/system_profile/v1.yaml
CodeHeeler: Removed min/max.  Validation failing on the invalids now.  Either a) int32 is unsigned by default, or b) integer before defaulted to unsigned, so now I need to update the invalids back to the original larger value we had for those.

CodeHeeler: int32 is [signed](https://swagger.io/docs/specification/data-models/data-types/) by default, but min/max must not be enforced as expected, adding them back resolved the failing tests.

kruai (approved):
Unfortunate that the min/max aren't enforced the way the docs imply, but oh well. Everything looks good to me!

thearifismail (commented):
LGTM

thearifismail (approved):

PR #93, opened 2022-08-25
Remove codeheeler from CODEOWNERS

kruai:
Dana is now part of the @RedHatInsights/host-based-inventory-committers group, so this removes the explicit reference in the CODEOWNERS file.


CodeHeeler (approved):
Thanks!

thearifismail (approved):

PR #94, opened 2022-09-09
[ESSNTL-2678] Create SAP object

CodeHeeler:
Create SAP object for consistency with other workloads and move existing SAP related fields to within it (sap_version, sap_system, sap_instance_number, and sap_sids if that one is not too complicated due to direct dependencies*).

*Included sap_sids in this draft, will reach out to folks and see what dependencies exist and how difficult they would be to resolve/time before merging this PR.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -406,6 +406,34 @@ $defs:
           type: string
           maxLength: 512
           example: "ex1, ex2, ex3"
+      sap:
+        description: Object containing data specific to the SAP workload
+        type: object
+        properties:
+          sap_system:
+            type: boolean
+            description: Indicates if SAP is installed on the system
+          sap_sids:
+            type: array
+            uniqueItems: true
+            items:
+              description: The SAP system ID (SID)
+              type: string
+              example: "H2O, ABC, XYZ"
+              maxLength: 3
+              pattern: '^[A-Z][A-Z0-9]{2}$'
+          sap_instance_number:
+            type: string
+            description: The instance number of the SAP HANA system (a two-digit number between 00 and 99)
+            example: '03, 05, 99'
+            maxLength: 2
+            pattern: '^[0-9]{2}$'
+          sap_version:
+            type: string
+            description: The version of the SAP HANA lifecycle management program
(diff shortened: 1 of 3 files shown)


CodeHeeler:
Flake8 doesn't like the syntax around the sap_sids, I'm not quite sure what needs to be done as I don't get these errors locally with my flake8 configured as it is.  Will discuss next week.

schemas/system_profile/v1.yaml
```diff
@@ -406,6 +406,34 @@ $defs:
...
+              example: "H2O, ABC, XYZ"
+              maxLength: 3
+              pattern: '^[A-Z][A-Z0-9]{2}$'
+          sap_instance_number:
+            type: string
+            description: The instance number of the SAP HANA system (a two-digit number between 00 and 99)
+            example: '03, 05, 99'
+            maxLength: 2
+            pattern: '^[0-9]{2}$'
+          sap_version:
```
kruai, line 431: Optional, just a thought... What do you think about removing the "sap_" prefix for these fields, since they're all under the "sap" object? I think it's fine either way; my brain just goes overdrive trying to reduce redundancy and such.
(Although we'd probably need to call the first one something besides just "system" for a boolean value to make sense)

CodeHeeler: It is less redundant within the object and looks like the other workloads similarly leave off the named prefix, so I'm ok with that.

Would it create more work for teams with dependencies than just accounting for the nested object without making name changes?

kruai: Good question. I'm assuming it'll be around the same amount of work because they're already changing the object structure, but I don't know for sure. I just want to make sure we nail down the names now, since changing them later will take way more effort. What do you think @beav @kahowell?

beav: if it's under the sap object already, :+1: to removing the prefixes

CodeHeeler: Ok, I'll remove the prefixes except for `sap_system` and push up those changes.

kruai (approved):

FabriciaDiniz (approved):
LGTM!

kruai:
@FabriciaDiniz it looks like your review didn't count because it's your @FabriciaDinizRH account that's in the GitHub group :)

FabriciaDinizRH (approved):

PR #97, opened 2022-10-31
add update_method to system_profile schema

thearifismail:
This PR adds update_method to system_profile_schema, [ESSNTL_3575](https://issues.redhat.com/browse/ESSNTL-3575)

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -570,7 +570,7 @@ $defs:
             example: '15.2.0, 12.5.3, 10.1.0'
             maxLength: 30
             x-wildcard: true
-      update_method:
+      system_update_method:
         description: System update method
         type: string
         maxLength: 10 # right a test with 11 chars
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -570,3 +570,9 @@ $defs:
             example: '15.2.0, 12.5.3, 10.1.0'
             maxLength: 30
             x-wildcard: true
+      update_method:
+        description: System update method
+        type: string
+        maxLength: 10 # right a test with 11 chars
+        enum: [dnf, rpm-ostree, yum]
+        example: "dnf, rpm-ostree, yum"
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -573,6 +573,6 @@ $defs:
       system_update_method:
         description: System update method
         type: string
-        maxLength: 10 # right a test with 11 chars
+        maxLength: 
         enum: [dnf, rpm-ostree, yum]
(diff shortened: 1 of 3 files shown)


FabriciaDinizRH (approved):

schemas/system_profile/v1.yaml
```diff
@@ -570,3 +570,9 @@ $defs:
             example: '15.2.0, 12.5.3, 10.1.0'
             maxLength: 30
             x-wildcard: true
+      update_method:
```
gmcculloug, line 573: @thearifismail I would suggest naming this `system_update_method` as it feels like a very generic name as it is now.  This would keep it inline with some of the other names like, `system_purpose` and `system_memory_bytes`.

schemas/system_profile/v1.yaml
```diff
@@ -570,3 +570,9 @@ $defs:
...
+        description: System update method
+        type: string
+        maxLength: 10 # right a test with 11 chars
```
CodeHeeler, line 576: Remove the comment on writing a test now that you're done. Your current invalids.py works as this is an enum, so anything too long will also not be one of the allowed values, but feel free to add another invalid value if you like.

schemas/system_profile/v1.yaml
```diff
@@ -570,3 +570,9 @@ $defs:
...
+        enum: [dnf, rpm-ostree, yum]
+        example: "dnf, rpm-ostree, yum"
```
CodeHeeler, line 578: I think the example should be one of the three, not all three, as that would fail being to long and not one of the enum options.

CodeHeeler: Although I see that we have this pattern of enumerating all options as the example in fields above, so maybe this should be a separate PR that changes each of those examples to correct examples.

CodeHeeler (changes requested):
Just a couple changes suggested, otherwise looks good.

thearifismail (commented):
responded to comments

CodeHeeler (approved):
Thanks for making the changes, lgtm.

schemas/system_profile/v1.yaml
```diff
@@ -570,3 +570,9 @@ $defs:
             example: '15.2.0, 12.5.3, 10.1.0'
             maxLength: 30
             x-wildcard: true
+      system_update_method:
+        description: System update method
+        type: string
+        maxLength:
```
kruai, line 576: Either give this a value or remove it

kruai (changes requested):

schemas/system_profile/v1.yaml
CodeHeeler: oh! the 10 was accidentally removed when the comment was removed.

kruai (approved):
Looks good now!

PR #99, opened 2022-11-03
feat: add releasever and basearch

vkrizan:
Adds `releasever` and `basearch` facts to system profile gathered from `yum_updates`.

This is to properly identify YUM repository URLs. Especially those that have `$releasever` and `$basearch` placeholders.

Note, that these two facts are already present in `yum_updates` JSON. However, the `yum_updates` requires additional parsing, hence this information should be globally present on system profile.

VULN-2455


strider:
👍

schemas/system_profile/v1.yaml
```diff
@@ -314,10 +314,20 @@ $defs:
...
+      releasever:
+        type: sting
+        description: Release name of the system distribution (from yum/dnf)
+        maxLength: 100
+        example: "7, 7.0, 7Server, 8, 8.4"
       arch:
         type: string
         maxLength: 50
         example: "ARM, x86_64, RISC-V"
+      basearch:
```
vkrizan, line 326: Looks like the `arch` should give the same result. `arch` comes from uname parser from some RPM package.

Let's also keep `basearch` as well to be sure that we use a proper value for repository URL substitutions.

schemas/system_profile/v1.yaml
```diff
@@ -314,10 +314,20 @@ $defs:
         description: The kernel version represented with a three, optionally four, number scheme.
         example: '4.18.2, 4.5.0, 5.1.0'
         x-wildcard: true
+      releasever:
```
thearifismail, line 317: Aren't max lengths for releasever and basearch too long and making it easy to introduce mistakes?

schemas/system_profile/v1.yaml
```diff
@@ -314,10 +314,20 @@ $defs:
...
+        type: sting
```
CodeHeeler, line 318: typo here, type string not sting, that will fix the failing validation test

CodeHeeler (changes requested):
Found the typo causing failing validation

CodeHeeler:
ping @gmcculloug to get another stakeholder's feedback on naming and inclusion of new fields on sp schema.

schemas/system_profile/v1.yaml
vkrizan: I've aligned the lengths to existing keys. `basearch` is aligned with `arch`s limit of 50, and `releasever` is aligned with `os_release`'s limit of 100.

CodeHeeler: Valid feedback, though the same could be said for `arch` above, so I would say not a blocker.

schemas/system_profile/v1.yaml
vkrizan: Fixed. Thank you!

thearifismail (approved):

CodeHeeler (approved):
LGTM, thanks!

CodeHeeler:
> @vkrizan Can you create a subtask on VULN-2404 similar to VULN-2447 to cover sync with hbi and xjoin? I'll create tasks for that in ESSNTL that you can link to like we did for mirrorlist.

ESSNTL-3672 & ESSNTL-3673 cover HBI and Xjoin

vkrizan:
> > @vkrizan Can you create a subtask on VULN-2404 similar to VULN-2447 to cover sync with hbi and xjoin? I'll create tasks for that in ESSNTL that you can link to like we did for mirrorlist.
> 
> ESSNTL-3672 & ESSNTL-3673 cover HBI and Xjoin

Done -> VULN-2461

kruai (approved):

schemas/system_profile/v1.yaml
kruai: @vkrizan Won't this effectively be duplicating the value? What makes `basearch` different from `arch`?

schemas/system_profile/v1.yaml
kruai: Why does this need a new field rather than using an existing one like `os_release`?
I'm also unclear on what values we can expect... the name implies "release version", but the description says it's the release name, but then the examples look like version numbers (except 7server).

kruai (commented):
I've got a couple of questions, but we should probably rename these fields as well, since their names are fairly unclear. They should follow the standard we use in the rest of our fields (`release_version` and `base_arch`, for instance)

thearifismail:
> I've got a couple of questions, but we should probably rename these fields as well, since their names are fairly unclear. They should follow the standard we use in the rest of our fields (`release_version` and `base_arch`, for instance)

@kruai I had the same question and also wanted to understand if the their values could be different from within a system, so I snooped around and found this is `/etc/yum.repos.d/fedora.repo`. I guessed it was better to keep the names the same as they appeared in URLs
```
metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
```

CodeHeeler:
> > I've got a couple of questions, but we should probably rename these fields as well, since their names are fairly unclear. They should follow the standard we use in the rest of our fields (`release_version` and `base_arch`, for instance)
> 
> @kruai I had the same question and also wanted to understand if the their values could be different from within a system, so I snooped around and found this is `/etc/yum.repos.d/fedora.repo`. I guessed it was better to keep the names the same as they appeared in URLs
> 
> ```
> metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
> ```

Yeah, I agreed with kruai's comment regarding the names not being in line with our usual expectations, but as vkrizan pointed out this was explicitly to provide the placeholders without further parsing of the yum repo json, and thearifismail's search corroborates this.

schemas/system_profile/v1.yaml
vkrizan: My understanding of the `arch` is that it provides architecture whilst `basearch` provides **family** of the architecture. `arch` comes from `uname.arch`, whilst `basearch` comes from DNF/YUM. The difference between `arch` and `basearch` is mentioned in [dnf documentation](https://dnf.readthedocs.io/en/latest/api_conf.html#dnf.conf.Conf.substitutions), and the mapping can be found in the [dnf repository](https://github.com/rpm-software-management/dnf/blob/96b08be69050cddbc4c3abdd227bd28969249db9/dnf/rpm/__init__.py#L88-L115). This can be compared with [`KNOWN_ARCHITECTURES` from Insights client](https://github.com/RedHatInsights/insights-core/blob/084e85d10f6e5df575427cf602807b7172ecae12/insights/parsers/installed_rpms.py#L84-L139).

schemas/system_profile/v1.yaml
vkrizan: `releasever` should be "release name of the system distribution" configured and/or detected by YUM/DNF. The value is affected by configuration and subscription manager, or otherwise detected. The detection is not easy to replicate, check implementation of [`detect_releasever` within dnf](https://github.com/rpm-software-management/dnf/blob/96b08be69050cddbc4c3abdd227bd28969249db9/dnf/rpm/__init__.py#L29-L70). Values depend on distribution and are created by Release Engineering.

On the other hand `os_release` is parsed out of `/etc/redhat-release` providing only a version part.

vkrizan:
For naming consistency with their source, I'd like to keep the names the same: `basearch` and `releasever`. That's how they are used and set within yum/dnf configuration.

kruai:
> Hello, is there more feedback?

As with the last PR, we need to run these by the stakeholders before we merge it. We have these meetings weekly, and the next one is today (in 25 mins). If nobody has any objections, I'll merge it

PR #101, opened 2022-12-05
[ci] Fix Python3.6 issue on Linux x86

strider:
These patches fix the following issue:

```
Error: Version 3.6 with arch x64 not found
```

Cause: there is no available build of Python 3.6 for ubuntu-22.04

ref: [Error: Version 3.6 with arch x64 not found](https://github.com/actions/setup-python/issues/544)

And also fixes the flake8 syntax errors.

Signed-off-by: Gael Chamoulaud (Strider) <gchamoul@redhat.com>

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -20,12 +20,12 @@ jobs:
     - run: spectral lint schemas/system_profile/v1.yaml
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python 3.6
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: 3.6
     - name: Install dependencies
--- a/tests/utils/invalids.py
+++ b/tests/utils/invalids.py
@@ -126,11 +126,11 @@ INVALID_SYSTEM_PROFILES = (
     {"network_interfaces": [{"ipv6_addresses": "0123:4567:89ab:cdef:0123:4567:89ab:cdef"}]},
     {"network_interfaces": [{"mtu": "15"}]},
     {"rhsm": {"version": "x" * 300}},
-    {"operating_system": {"name": "RHEL"}}, # Incomplete OS definition
-    {"operating_system": {"name": "RHEL", "major": 9}}, # Incomplete OS definition
-    {"operating_system": {"major": 8, "minor": 7}}, # Incomplete OS definition
-    {"operating_system": {"name": "ABCD", "major": 9, "minor": 0}}, # Invalid name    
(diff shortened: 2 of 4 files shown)


kruai (approved):

FabriciaDinizRH (approved):

.github/workflows/pr.yml
```diff
@@ -20,12 +20,12 @@ jobs:
     - run: spectral lint schemas/system_profile/v1.yaml
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
thearifismail, line 23: If possible, it be better to use the `latest`, which means one less thing to worry about.

tests/utils/invalids.py
```diff
@@ -126,11 +126,11 @@
     {"network_interfaces": [{"ipv6_addresses": "0123:4567:89ab:cdef:0123:4567:89ab:cdef"}]},
     {"network_interfaces": [{"mtu": "15"}]},
     {"rhsm": {"version": "x" * 300}},
-    {"operating_system": {"name": "RHEL"}}, # Incomplete OS definition
-    {"operating_system": {"name": "RHEL", "major": 9}}, # Incomplete OS definition
-    {"operating_system": {"major": 8, "minor": 7}}, # Incomplete OS definition
-    {"operating_system": {"name": "ABCD", "major": 9, "minor": 0}}, # Invalid name    
-    {"operating_system": {"name": "RHEL", "major": "9", "minor": "0"}}, # Invalid types
+    {"operating_system": {"name": "RHEL"}},  # Incomplete OS definition
```
thearifismail, line 129: Are these changes added by some linter?

thearifismail (approved):
Approving with minor comments.

.github/workflows/pr.yml
strider: There is no Python3.6 on ubuntu-latest anymore, that's why we have to use 20.04.
Or drop python3.6 support and test against 3.7/3.8/3.9.

tests/utils/invalids.py
strider: Fixed manually by myself.

See https://github.com/RedHatInsights/inventory-schemas/actions/runs/3620380106/jobs/6102532082#step:5:1

.github/workflows/pr.yml
thearifismail: Okay that make sense.  Unless the build node requires Python 3.6, it should be dropped.

CodeHeeler (approved):
LGTM, though I agree we should at some point drop support of the old version if it's not required by anything and stick with `latest` again in the future.

strider:
Well, doesn't make sense to wait ... :D

strider:
Yes, it actually makes sense to do that in another PR as it implies bumping dependencies.

strider:
So this PR could be merged as it for now.

PR #103, opened 2023-01-17
Add test field to trigger SP validator

kruai:
We need to trigger the SP Validator in the Prod environment to verify that it works with its current Kafka consumer group-id config.


kruai:
It triggered the 12pm EST job, which failed due to the group-id config. I'll make a PR fixing that for both the sp-validator job and the host-synchronizer job.

PR #110, opened 2023-06-02
Added systemd state object to the system_profile inventory schema.

roliveri:
Addresses: https://issues.redhat.com/browse/RHINENG-484

Added systemd state object to the system_profile inventory schema.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -291,12 +291,12 @@ $defs:
             enum: [initializing, starting, running, degraded, maintenance, stopping]
             example: 'initializing, starting, running, degraded, maintenance, stopping'
           jobs_queued:
-            description: The nomber of jobs jobs_queued
+            description: The number of jobs jobs_queued
             example: 4
             type: integer
             minimum: 0
           failed:
-            description: The nomber of jobs failed
+            description: The number of jobs failed
             example: 1
             type: integer
             minimum: 0
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -276,6 +276,30 @@ $defs:
           type: string
           maxLength: 30
           example: "ex1, ex2, ex3"
+      systemd:
+        description: Object for whole system systemd state, as reported by systemctl status --all
+        type: object
+        required:
+          - state
+          - jobs_queued
+          - failed
+        properties:
+          state:
+            description: The state of the systemd subsystem
+            type: string
+            maxLength: 12
+            enum: [initializing, starting, running, degraded, maintenance, stopping]
+            example: 'initializing, starting, running, degraded, maintenance, stopping'
+          jobs_queued:
+            description: The nomber of jobs jobs_queued
+            example: 4
+            type: integer
+            minimum: 0
+          failed:
+            description: The nomber of jobs failed
+            example: 1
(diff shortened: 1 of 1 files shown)


schemas/system_profile/v1.yaml
```diff
@@ -276,6 +276,30 @@ $defs:
...
+            maxLength: 12
+            enum: [initializing, starting, running, degraded, maintenance, stopping]
+            example: 'initializing, starting, running, degraded, maintenance, stopping'
+          jobs_queued:
+            description: The nomber of jobs jobs_queued
+            example: 4
+            type: integer
+            minimum: 0
+          failed:
+            description: The nomber of jobs failed
```
thearifismail, line 299: number

schemas/system_profile/v1.yaml
```diff
@@ -276,6 +276,30 @@ $defs:
...
+          - failed
+        properties:
+          state:
+            description: The state of the systemd subsystem
+            type: string
+            maxLength: 12
+            enum: [initializing, starting, running, degraded, maintenance, stopping]
+            example: 'initializing, starting, running, degraded, maintenance, stopping'
+          jobs_queued:
+            description: The nomber of jobs jobs_queued
```
thearifismail, line 294: number

schemas/system_profile/v1.yaml
```diff
@@ -276,6 +276,30 @@ $defs:
...
+        required:
+          - state
+          - jobs_queued
+          - failed
+        properties:
+          state:
+            description: The state of the systemd subsystem
+            type: string
+            maxLength: 12
+            enum: [initializing, starting, running, degraded, maintenance, stopping]
```
thearifismail, line 291: Is there a steady state?  What about updated or good-state

thearifismail (commented):
What is the use case?  When done, will the system_profile include the whole output of `systemctl status --all`?
The `systemd` requires `state`, `job_queued`, or `failed` but required where?

schemas/system_profile/v1.yaml
roliveri: The use case is described in:
https://issues.redhat.com/browse/RHINENG-484
"Policies need to use systemd state so that customers can create custom rule/policy."

It doesn't contain the whole output of `systemctl status --all`, it filters the first 3 lines:
https://issues.redhat.com/browse/INSPEC-401

If the systemd object is in the system_profile, those fields are required - must be defined in the object.

schemas/system_profile/v1.yaml
roliveri: I would think the steady state would be "running".

roliveri:
@kruai Are those dippy-bot failures a concern? I'm not sure how to determine the cause of the failures.

kruai:
> @kruai Are those dippy-bot failures a concern? I'm not sure how to determine the cause of the failures.

The main thing I look out for in these PRs is whether the number of failures is higher for the PR's spec vs the existing spec. It's not increasing for this PR, so it's unlikely to introduce new issues.

Now, it's certainly a bit concerning that "rhsm-system-profile-bridge" seems to have a lot of failures sometimes, but that doesn't seem related to the PR.

thearifismail:
LGTM

thearifismail (approved):
LGTM

PR #111, opened 2023-06-12
Add cloud external IPv4

lzap:
Hello,

newly introduced services allows customers to build RHEL image, upload it to major cloud providers and then launch them directly from our portal. We want customers to be able to also SSH to the machine without leaving our portal. In order to do that, they need to know the public IP address also known as external IP address in some hyperscalers.

Insights currently contains IP addresses from the system itself, but in IaaS public IP addresses are not assigned directly to the instances, they are NATed by the cloud infrastructure. Therefore, new specs were added to the insights-core repository for all three major providers via their respective metadata APIs: https://github.com/RedHatInsights/insights-core/pulls?q=is%3Apr+is%3Aclosed+author%3Alzap

This patch adds a new field called `cloud_external_ipv4_addresses`, little longer to my taste but this nicely goes with existing field named `cloud_provider`. Some clouds to provide both IPv4 and IPv6 so it is useful to have this in the name. Finally, some clouds also allows you to assign multiple external addresses to a single instance. Thus it is an array.

I am completely shooting in the dark here, I haven’t tested the patch and no idea what to do next. Please help me with this, thank you!


kruai:
@lzap I will review this PR, but please address the linting issues first. Thanks!

kruai:
@lzap I had a meeting with the other System Profile stakeholders, and they prefer something closer to your original naming. Would you be all right calling it `public_ipv4_addresses`?

lzap:
Sure, renamed, fixed linter complaints.

lzap:
Done, can you help me what to do next in order to put the new data to inventory page?

kruai (approved):

PR #115, opened 2023-08-08
RHINENG-1494 adding proposed changes to os name

kami-nashi:
Currently, in the system_profile schema, under operating_system:properties:name the schema is defined for [RHEL, CentOS]. During a previous conversation it was mentioned that CentOS 7 systems need to say "CentOS Linux", and if we ever enabled CentOS Stream, that name would need added as well. With my proposed change to puptoo, systems fail to update/check in with the new functionality of the "CentOS Linux" name because the schema does not allow it.  I am proposing that we simply update that one name.

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -324,8 +324,8 @@ $defs:
             description: Name of the distro/os
             type: string
             maxLength: 6
-            enum: [RHEL, CentOS]
-            example: "RHEL, CentOS, RHEL"
+            enum: [RHEL, CentOS, CentOS Linux]
+            example: "RHEL, CentOS, CentOS Linux, RHEL"
       os_release:
         type: string
         maxLength: 100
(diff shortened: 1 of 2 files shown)


kami-nashi:
leaving comment to connect to RHINENG-1494

kami-nashi:
will reach out later today to discuss the changes

schemas/system_profile/v1.yaml
```diff
@@ -324,8 +324,8 @@ $defs:
             description: Name of the distro/os
             type: string
             maxLength: 6
-            enum: [RHEL, CentOS]
-            example: "RHEL, CentOS, RHEL"
+            enum: [RHEL, CentOS, CentOS Linux]
+            example: "RHEL, CentOS, CentOS Linux, RHEL"
```
kruai, line 328:
```suggestion
            example: "RHEL, CentOS, CentOS Linux"
```

kruai (approved):

PR #116, opened 2023-10-04
ci(RHINENG-1493): Fix spectral linting issue

strider:

PR #117, opened 2023-10-05
RHINENG-2402 add failed services for systemd

r14chandra:
[RHINENG-2402](https://issues.redhat.com/browse/RHINENG-2402)

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -300,6 +300,14 @@ $defs:
             example: 1
             type: integer
             minimum: 0
+          failed_services:
+            description: List of all failed jobs.
+            type: array
+            items:
+              description: The name of a failed job.
+              maxLength: 1000
+              type: string
+              example: "ex1, ex2, ex3"
       operating_system:
         description: Object for OS details. Supports range operations
         type: object


strider:
Hi @r14chandra,

You can safely ignore the validations linting issue, which is not related to your changes. FYI, I am trying to fix it with #116.

strider (approved):

kruai:
@r14chandra With the addition of this field, do we still need `failed`, the field directly above this one?

r14chandra:
> @r14chandra With the addition of this field, do we still need `failed`, the field directly above this one?

@kruai Yes we require that as well because `failed` will just show the total no. of failed units.

kruai:
> > @r14chandra With the addition of this field, do we still need `failed`, the field directly above this one?
> 
> @kruai Yes we require that as well because `failed` will just show the total no. of failed units.

I figured, but it seems redundant since you'll also have an array of all the services, which you could just get the length/size of

r14chandra:
> > > @r14chandra With the addition of this field, do we still need `failed`, the field directly above this one?
> > 
> > 
> > @kruai Yes we require that as well because `failed` will just show the total no. of failed units.
> 
> I figured, but it seems redundant since you'll also have an array of all the services, which you could just get the length/size of

Yes, correct. But I don't know if the team requesting this feature would want to discard the use of `failed`.

kruai (approved):

PR #121, opened 2023-11-13
DISCOVERY-325: Add virtual_host_uuid

bruno-fs:
virtual_host_uuid field is added to reflect the hypervisor identity (subiscription_manager_id) of a system's hypervisor (which is typically collected by virt-who).

https://issues.redhat.com/browse/DISCOVERY-325

### DIFF
--- a/schemas/system_profile/v1.yaml
+++ b/schemas/system_profile/v1.yaml
@@ -650,3 +650,9 @@ $defs:
         maxLength: 10
         enum: [dnf, rpm-ostree, yum]
         example: "dnf, rpm-ostree, yum"
+      virtual_host_uuid:
+        description: Hypervisor host identity (subscription manager id)
+        type: string
+        pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
+        maxLength: 36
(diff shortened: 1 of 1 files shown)


schemas/system_profile/v1.yaml
```diff
@@ -650,3 +650,9 @@ $defs:
         maxLength: 10
         enum: [dnf, rpm-ostree, yum]
         example: "dnf, rpm-ostree, yum"
+      virtual_host_uuid:
+        description: hypervisor host identity (subscription manager id)
+        type: string
+        pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
```
infinitewarp, line 656: Looking elsewhere in this schema file, match patterns for hex numbers are somewhat inconsistent (MAC addresses match on both cases, but other UUIDs only check lowercase), but my paranoid side thinks the UUID patterns should include matches on uppercase even though I don't know if they will ever be generated that way. 🤷

```suggestion
        pattern: '[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
```

infinitewarp: As discussed on Slack, let's just remain consistent with the other UUID patterns and expect the reporters to set the format consistently.

kruai:
Hmm, dippy-bot leaving multiple comments in a row is a new one. I'll look into that.
Don't worry about the linting issues - that's being investigated

kruai (approved):

PR #128, opened 2024-04-08
RHINENG-9246: add "cached_update_image_digest"

JoySnow:
Related `system_profile` changes at https://github.com/RedHatInsights/insights-puptoo/pull/328


JoySnow:
Close this PR here, as the current system_profile `bootc_status.cachedUpdate` in schema works for RHINENG-9246 and no need to introduce a schema change.

PR #134, opened 2024-05-31
RHINENG-10250: add "crowdstrike" field to schema

JoySnow:
[RHINENG-10250](https://issues.redhat.com/browse/RHINENG-10250)


schemas/system_profile/v1.yaml
```diff
@@ -780,3 +780,17 @@ $defs:
             type: boolean
             x-indexed: false
             example: true
+      crowdstrike:
```
thearifismail, line 783: What is the purpose of adding `crowdstrike` to system profile? Is it to identify the system, i.e. what does this system do?  If the answer is "yes", then how about putting `crowdstriker` in a new property, say `purpose`.  

The reason for this suggestion is that `system_profile` column is already too big and adding new properties/attributes to system_profile has to the potential to grow too big and harder to maintain.  Furthermore, I see new work on the horizon that will add more systems types to inventory and those systems  can be put under the same property/section as "crowdstrike"

JoySnow: Hey @thearifismail , your consideration totally makes sense. I agree we can put `crowdstrike` inside a more general property. 

Basically, CrowdStrike is more about a security service instead of a workload (like intersystem, sap ...) in a system. Based on this, I am not sure if `purpose` is the best choice. What about some name like `third_party_services`?  

Feel free to let me know your suggestion/though.

JoySnow:
@thearifismail ,  I updated the code to wrap this "crowdstrike" into a new attribute `third_party_services ` based on [our discussion above](https://github.com/RedHatInsights/inventory-schemas/pull/134#discussion_r1624199312) .  

Feel free to update to other meaningful attribute name based on your practice.

thearifismail (approved):
LGTM

schemas/system_profile/v1.yaml
thearifismail: @joy I am going rescind the approval because @kruai thinks that it should be discussed with the users of inventory-schema

thearifismail (commented):
This should be discussed by all the users of Inventory Schema

kruai (approved):
Folks are on board with the `third_party_services` approach, so I'm approving.

PR #140, opened 2024-09-25
RHINENG-12955: add image builder fields to schema

croissanne:


fstavela (approved):

schemas/system_profile/v1.yaml
```diff
@@ -816,4 +816,20 @@ $defs:
                 description: CrowdStrike running Falcon Sensor version
                 type: string
                 example: "7.14.16703.0, 6.33.13003.0"
-                maxLength: 32
\ No newline at end of file
+                maxLength: 32
+      image_builder:
```
thearifismail, line 820: @croissanne Got some questions:
1. What services are the primary consumers of this information?
2. Will queries be run to find hosts using "compliance_policy_id" and "compliance_prorifle_id"?
3. If only one service will be using ImageBuidler, then could this information be written in a files and copied to the image?

schemas/system_profile/v1.yaml
```diff
@@ -816,4 +816,20 @@ $defs:
...
+        description: Object containing image builder facts
+        type: object
+        properties:
```
thearifismail, line 823: Shouldn't one of the properties be a required property?

schemas/system_profile/v1.yaml
skateman: 1. ours (compliance)
2. yes (on cyndi)
3. I have no ide about this

In general the idea is that compliance should be able to search for the systems that have been built with ImageBuilder and are hardened against either one of our policies or a SCAP profile.

thearifismail: What does compliance do upon finding out a host built using image builder and hardened against some policy?

skateman: When a user creates/edits a policy, we want to offer the user something like "hey dude, there's a bunch of hosts built with this SCAP profile, do you want to add them to this policy?" on the other hand we want to tap into the kafka events and catch any new systems that boot up for the first time and auto-register them to compliance (if the user wants).

schemas/system_profile/v1.yaml
croissanne: Neither of these are required, we might add more fields in future however.

schemas/system_profile/v1.yaml
croissanne: I'm not entirely sure what you mean by 3, we write the facts into the image when we build it, we'd like to have these facts exposed in the inventory so that compliance can make use of them.

thearifismail: Question #3 means, if `system_profile.image_builder.compliance_policy_id` and `system_profile.image_builder.compliance_profile_id` are not available in `hosts` table, can this information be acquired in any other way?

schemas/system_profile/v1.yaml
```diff
@@ -816,4 +816,20 @@ $defs:
...
+          compliance_policy_id:
+            description: The compliance policy that was used and applied during the image build
+            type: string
+            example: 89b52baa-9912-4edc-9ed5-be15c06eaaa9
+            pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
+            maxLength: 36
+          compliance_profile_id:
```
thearifismail, line 830: Both properties are `ids`, but one's value is of type UUID and other plain string.  Shouldn't both be UUIDs and particularly of type UUID, so that they could be validated.

tests/utils/invalids.py
```diff
@@ -396,5 +396,17 @@
         "crowdstrike": {  # Must be a string, not a boolean
             "falcon_version": True,
         }
+    }},
+    {"image_builder": { # Must be a string
+        "compliance_policy_id": 10,
```
thearifismail, line 401: Schema definition states `pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'` MUST be followed but the test uses a values of "10". and "some_profile_id".

schemas/system_profile/v1.yaml
thearifismail: if this the case, please set type for each property so that their values could be validated to prevent bogus data getting in and make it secure.

thearifismail (approved):
Rescinding approval and seeking answer to comments.

schemas/system_profile/v1.yaml
skateman: They cannot be retrieved any other way from compliance :/ it has to be syndicated via cyndi as well. And yes, in their own context they are both IDs but different kid. One of them should be a UUID and the other one an OpenSCAP id, this is on purpose.

thearifismail: Thanks.   Got it.

thearifismail (approved):

tests/utils/invalids.py
croissanne: I'm confused, the profile_id is just a string, so "some_profile_id" is correct. The compliance_policy_id is a uuid.

schemas/system_profile/v1.yaml
croissanne: Depends, like you could get it from the blueprint in the image builder database but it's hard to tie the host to a blueprint (we want to add a fact to the schema for this as well, so you could theoretically go from a system page to image builder and highlight the blueprint). And you could maybe infer it by booting the system and poking around.

But there is no easy way for compliance to get the IDs for compliance without these facts.

schemas/system_profile/v1.yaml
croissanne: Do you want to have a more specific pattern? I'm not sure if there are restrictions for openscap patterns @skateman  do you know? Like lowercase characters, dots and underscores?

skateman: image
In our case it starts with `xccdf_org.ssgproject.content_profile`

croissanne:
@thearifismail do you need more info or is this good to go?

croissanne:
I can't actually press the merge button, can someone with rights do it?

thearifismail:
Done!!!

skateman:
@thearifismail what else is needed to have this on stage working?

PR #145, opened 2024-12-04
CHAINSAW-234 Add rhsm.environments to schema

Tomboyo:
Adds the rhsm.environments field to the schema. See associated change to insights-host-inventory, https://github.com/RedHatInsights/insights-host-inventory/pull/2112, which reflects these changes there per the contrib section of the README.


kruai (changes requested):
@Tomboyo Looks good, but can we set a max length for the field? I was going to suggest 32 because of the examples, but one of your tests in `valids` has a longer string

thearifismail (approved):

PR #163, opened 2025-08-06
update captured-date field to be datetime

jpramos123:
Update `captured_date` field to be a date-time field.


rodrigonull (approved):
LGTM

PR #164, opened 2025-08-08
chore(RHINENG-19793) Add x-dynamic to fields that will be in the dynamic table

kruai:
This PR addresses [RHINENG-19793](https://issues.redhat.com/browse/RHINENG-19793). Adds the `x-dynamic` identifier to fields that are in the `system_profiles_dynamic` table. This will allow HBI to determine the correct table association for each field.


ezr-ondrej (approved):
🧡 it, thank you! 👍

kruai:
This just adds an extra unused property to some fields and does not affect the functionality at all, so this should be safe to merge without waiting on the validator.

